PR #1, opened 2018-08-08
Vote Graph and GHOST-hunting

rphmeier:
This PR contains the logic for imposing a further DAG on top of the blockchain, where nodes are blocks which have been voted on directly.

Cumulative votes (flowing back to the "base" -- in practice, the last finalized block) are kept updated, and a "GHOST" calculation can be done on this DAG to find the highest block referenced in a node or edge of the DAG that fulfills some vote-predicate. This block may fall between some vote-nodes.

The GHOST logic, as it stands right now, operates under the assumption that the predicate will evaluate to `true` for at most one branch of any fork. Full GHOST is more expensive to implement as we have to compare the cumulative scores of all children and can't early-exit upon finding the first child which satisfies the predicate.


src/lib.rs
```diff
@@ -0,0 +1,67 @@
+// Copyright 2018 Parity Technologies (UK) Ltd.
+// This file is part of finality-afg.
+
+// Polkadot is free software: you can redistribute it and/or modify
```
gnunicorn: Shouldn't this be `finality-afg` rather than `Polkadot`?

src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Polkadot is distributed in the hope that it will be useful,
```
gnunicorn: here, too.

src/vote_graph.rs
```diff
@@ -0,0 +1,515 @@
...
+			base: base_hash,
+		}
+	}
+
+	/// Insert a vote with given value into the graph at given hash and number.
+	pub fn insert<C: Chain<H>>(&mut self, hash: H, number: usize, vote: V, chain: &C) -> Result<(), Error> {
+		match self.find_containing_nodes(hash.clone(), number) {
+			Some(containing) => if containing.is_empty() {
+				self.append(hash.clone(), number, chain)?;
+			} else {
```
gnunicorn: Shouldn't this be indented one level more rather than be on the level as the `match`-arms?

src/vote_graph.rs
```diff
@@ -0,0 +1,515 @@
...
+		// we have a frontier of vote-nodes which individually don't have enough votes
+		// to pass the threshold but some subset of them join either at `active_node` or at some
+		// descendent block of it, giving that block sufficient votes.
+		Some(self.ghost_find_merge_point(node_key, active_node, condition))
+	}
+
+	// given a key, node pair (which must correspond), assuming this node fulfills the condition,
+	// this function will find the highest point at which its descendents merge, which may be the
+	// node itself.
+	fn ghost_find_merge_point<'a, F>(&'a self, mut node_key: H, mut active_node: &'a Entry<H, V>, condition: F) -> (H, usize)
```
gnunicorn: please wrap.

src/vote_graph.rs
```diff
@@ -0,0 +1,515 @@
...
+		// TODO: for long ranges of blocks this could get inefficient
+		for offset in 1usize.. {
+			let mut new_best = None;
+			for d_node in descendent_nodes.iter() {
+				if let Some(d_block) = d_node.ancestor_block(base_number + offset) {
+					match descendent_blocks.binary_search_by_key(&d_block, |&(ref x, _)| x) {
+						Ok(idx) => {
+							descendent_blocks[idx].1 += d_node.cumulative_vote.clone();
+							if condition(&descendent_blocks[idx].1) {
+								new_best = Some(d_block.clone());
```
gnunicorn: uui.... 7 indentation levels... I can still follow, but it is a bit against the style guide ;)

andresilva:
I'm still reviewing this, please don't merge it yet.

src/vote_graph.rs
```diff
@@ -0,0 +1,515 @@
...
+
+		let base_number = active_node.number;
+		let (mut best_hash, mut best_number) = (node_key, active_node.number);
+		let mut descendent_blocks = Vec::with_capacity(descendent_nodes.len());
+
+		// TODO: for long ranges of blocks this could get inefficient
+		for offset in 1usize.. {
+			let mut new_best = None;
+			for d_node in descendent_nodes.iter() {
+				if let Some(d_block) = d_node.ancestor_block(base_number + offset) {
```
andresilva: We return `None` if the block is past the parent vote-node of `d_node`. It's ok to ignore it since the parent vote-node of `d_node` should eventually be reached (if needed) as the offset grows.

src/vote_graph.rs
```diff
@@ -0,0 +1,515 @@
...
+				Some(new_best) => {
+					best_hash = new_best;
+					best_number += 1;
+				}
+				None => break,
+			}
+
+			descendent_blocks.clear();
+			descendent_nodes.retain(
+				|n| n.in_direct_ancestry(&best_hash, best_number).unwrap_or(false)
```
andresilva: Shouldn't this be `unwrap_or(true)`? We're removing nodes that are too high compared to the current best GHOST node, i.e. there are other vote nodes in the middle. It doesn't necessarily mean that they're not part of this branch.

andresilva:
Really clean code :+1: It's possible I misunderstood some parts of how this works or what invariants hold.

src/vote_graph.rs
```diff
@@ -0,0 +1,515 @@
...
+		Ok(())
+	}
+
+	/// Find the best GHOST descendent of the given block.
+	/// Pass a closure used to evaluate the cumulative vote value.
+	///
+	/// The GHOST (hash, number) returned will be the block with highest number for which the
+	/// cumulative votes of descendents and itself causes the closure to evaluate to true.
+	///
+	/// This assumes that the evaluation closure is one which returns true for at most a single
```
andresilva: This holds assuming an honest majority will follow the heaviest chain.

src/vote_graph.rs
```diff
@@ -0,0 +1,515 @@
...
+	// get ancestor vote-node.
+	fn ancestor_node(&self) -> Option<H> {
+		self.ancestors.last().map(|x| x.clone())
+	}
+}
+
+/// Maintains a DAG of blocks in the chain which have votes attached to them,
+/// and vote data which is accumulated along edges.
+pub struct VoteGraph<H: Hash + Eq, V> {
+	entries: HashMap<H, Entry<H, V>>,
```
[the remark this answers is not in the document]
rphmeier: true. I think it's fair to treat hashes as unique

src/vote_graph.rs
```diff
@@ -0,0 +1,515 @@
...
+				});
+
+				new_entry.descendents.push(descendent);
+				new_entry.cumulative_vote += entry.cumulative_vote.clone();
+			}
+
+			maybe_entry
+		});
+
+		if let Some(new_entry) = produced_entry {
```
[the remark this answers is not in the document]
rphmeier: indeed. but what am I supposed to do, panic?

src/vote_graph.rs
rphmeier: If we get `None` back, then we've exhausted all the blocks in the "edge" between our GHOST vote-node and the descendent. 

Since we're looking for an ancestor-block of the descendent vote-node's block which has enough cumulative votes, once we've reached this point we know that it's safe to remove the descendent node from our frontier, as we've checked all ancestors of the descendent vote-node already.

The other theoretical case (which cannot occur in this part of the code) where `in_direct_ancestry` returns `None` is where the block is _earlier_ than the best GHOST vote-node. But we don't care about that case anyway since we're only looking for descendent blocks of that node.

src/vote_graph.rs
rphmeier: yep, it's not ideal but I couldn't find a way to rewrite.

src/vote_graph.rs
rphmeier: the parent vote-node of `d_node` will always be the GHOST vote-node. `ancestor_block` will return `None` when we have reached the block corresponding to `d_node`, at which point we will want to prune anyway.

src/vote_graph.rs
```diff
@@ -0,0 +1,515 @@
...
+				None => Some(hash),
+				Some(ref x) if !x.is_empty() => Some(x[0].clone()),
+				Some(_) => None,
+			})
+			.and_then(|hash| get_node(&hash).ancestor_node())
+			.unwrap_or_else(|| self.base.clone());
+
+		let mut active_node = get_node(&node_key);
+
+		if !condition(&active_node.cumulative_vote) { return None }
```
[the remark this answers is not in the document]
rphmeier: this function is defined to always return with a descendent of the supplied node. this question cuts to the crux of the issue mentioned in the PR description and comment on this function: that `find_ghost` isn't a proper GHOST but a weird "majority-GHOST" thing where you either have one descendent of a node that meets the condition or none. So we find the best block that meets the condition or none and we don't need to backtrack because of the majority condition.

src/vote_graph.rs
rphmeier: I don't think so but open for discussion

src/vote_graph.rs
rphmeier: I think it holds as long as nobody double-votes (or in the case of supermajority, less than 1/3 double-vote). But I intend to filter out all double-votes in a higher level that instantiates the graph in a follow-up PR, so the property should always hold in our use-case.

I would like more unit tests to ensure that my logic is sound here

src/vote_graph.rs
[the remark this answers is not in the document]
rphmeier: to note why the property of returning a descendent is useful: this function is needed for the vote-casting logic, where we want to cast votes either after some time has passed _or_ as soon as `GHOST(S) == GHOST(S + Y)` where `Y` could be any new valid votes to add to the vote set. 

And as the document shows, the GHOST block never backtracks as we include more votes. So by bookkeeping the current `GHOST(S)` node we can easily check `GHOST(S + Y)` with another call to the same function. Although it can still be optimized a bit it should still be pretty fast.

src/vote_graph.rs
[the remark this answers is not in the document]
andresilva: No, I think it's fine as is, it was just a comment regarding my understanding.

PR #5, opened 2018-08-27
Handle equivocations correctly

### LINKED ISSUE #4: Count weight from equivocations
Based on https://github.com/paritytech/finality-afg/pull/2#pullrequestreview-149297834

But only in vote-nodes where no vote-weight from the equivocating node has already been counted.
As a matter of practicality, we can rewrite the `Vote` object used in the round counter as 
```rust
enum Kind {
    Prevote,
    Precommit,
}

// better name than `Vote`
struct Weight<Id> {
    prevote: usize,
    precommit: usize,
    votes: HashMap<(Id, Kind), usize>,
}
```

and alter the `Add` operation to only add weight where the voter hasn't been counted yet.
This will probably impede performance somewhat because we will have to do several hash-map queries per addition.

---- 
cc @AlistairStewart

Another open question is whether this introduces a DoS vector, as we could be spammed with votes on all possible blocks by an equivocating authority.

We discussed a few simple rules for dropping/forgetting votes, but we may want to investigate more:
  - The vote is on the same block as another (in this case, the equivocation is in the signature)
  - The vote is not on a descendent of the last finalized block of the last round.

rphmeier:
Closes #4

This introduces a `bitfield` module, which is used to lazily attach bitfields for equivocating (double-voting) validators as soon as we have an equivocating validator. The bitfields are uninitialized in the general case and are fast. The node-weight addition logic has been altered to avoid double-counting any equivocated votes.

The second, and probably most error-prone thing this does is to alter the logic for determining if a round is completable. The completable check sometimes needs to determine if any descendents of a block could possibly get 2/3+ vote-weight, and this PR alters the logic there to account for possible equivocations on blocks (i.e. not discounting any vote-weight which has already been spent).

### DIFF
--- a/src/bitfield.rs
+++ b/src/bitfield.rs
@@ -175,11 +175,13 @@ impl<Id: Eq> LiveBitfield<Id> {
 		let word_off = bit_idx / 64;
 		let bit_off = bit_idx % 64;
 
-		// TODO: if this isn't `Some`, something has gone really wrong.
-		// log it?
+		// If this isn't `Some`, something has gone really wrong.
 		if let Some(word) = self.bits.get_mut(word_off) {
 			// set bit starting from left.
 			*word |= 1 << (63 - bit_off)
+		} else {
+			warn!(target: "afg", "Could not set bit {}. Bitfield was meant to have 2 bits for each of {} validators.",
+				bit_idx, self.shared.n_validators);
 		}
 	}
 
--- a/src/bitfield.rs
+++ b/src/bitfield.rs
@@ -97,6 +97,24 @@ impl<Id: Eq> Bitfield<Id> {
 			}
 		}
 	}
+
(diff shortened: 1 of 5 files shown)


src/bitfield.rs
```diff
@@ -0,0 +1,396 @@
...
+		let val_off = self.shared.get_or_register_equivocator(id, weight)?;
+		self.set_bit(val_off * 2 + 1);
+		Ok(())
+	}
+
+	fn set_bit(&mut self, bit_idx: usize) {
+		let word_off = bit_idx / 64;
+		let bit_off = bit_idx % 64;
+
+		// TODO: if this isn't `Some`, something has gone really wrong.
```
gnunicorn: yes, please log this with warning or higher. I don't think it can happen at the moment, but IMHO stuff in here should scream loudly if it assumptions are not met.

PR #7, opened 2018-08-28
fix finality logic

rphmeier:
Previously was just returning any block with threshold precommits. Now ensures it has both prevotes and precommits. The test is fixed as well.


rphmeier:
The `update_trackers` function is getting a bit unwieldy, might be worth a refactor soon.

PR #8, opened 2018-09-10
Voting + round-change

rphmeier:
This pull request implements the actual algorithm of the finality gadget, including choosing when to start rounds, how to react to messages, choosing blocks to cast votes on, and producing finality notifications.

Most of the changes here is the introduction of a `Voter` future, which relies on an `Environment` implementation for setting up timers and networking and to pass finality information to.

The `Voter` has a `best_round` and some `background_rounds` which are prior. Background rounds are kept and continued to be polled until a block with the same number as their round-estimate is finalized. This is safe because the round-estimate is a calculation of the best possible block that could be finalized in that round -- once we've finalized something equal to or beyond that point, the round becomes irrelevant.

This does pass on a lot of the work of correct communication to the `Environment` layer: all signing, message routing, and buffering of incoming messages is left outside of this crate for now (although creating some common utilities for that) may be useful in the future.

### DIFF
--- a/src/voter.rs
+++ b/src/voter.rs
@@ -153,10 +153,10 @@ pub struct VotingRound<H, E: Environment<H>> where H: Hash + Clone + Eq + Ord + 
 	votes: Round<E::Id, H, E::Signature>,
 	incoming: E::In,
 	outgoing: Buffered<E::Out>,
-	state: Option<State<E::Timer>>,
-	bridged_round_state: Option<::bridge_state::PriorView<H>>,
-	last_round_state: ::bridge_state::LatterView<H>,
-	primary_block: Option<(H, usize)>,
+	state: Option<State<E::Timer>>, // state machine driving votes.
+	bridged_round_state: Option<::bridge_state::PriorView<H>>, // updates to later round
+	last_round_state: ::bridge_state::LatterView<H>, // updates from prior round
+	primary_block: Option<(H, usize)>, // a block posted by primary as a hint. TODO: implement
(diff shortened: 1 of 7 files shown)


src/voter.rs
```diff
@@ -0,0 +1,614 @@
...
+		let last_round_estimate = last_round_state.estimate.clone()
+			.expect("Rounds only started when prior round completable; qed");
+
+		let find_descendent_of = match self.primary_block {
+			None => {
+				// vote for best chain containing prior round-estimate.
+				last_round_estimate.0
+			}
+			Some(ref primary_block) => {
+				// we will vote for the best chain containing `p_hash` iff
```
gnunicorn: typo `iff`

src/voter.rs
```diff
@@ -0,0 +1,614 @@
...
+	fn notify(&self, last_state: RoundState<H>, new_state: RoundState<H>) {
+		if last_state == new_state { return }
+
+		if let Some(ref b) = self.bridged_round_state {
+			b.update(new_state.clone());
+		}
+
+		if last_state.finalized != new_state.finalized && new_state.completable {
+			// send notification only when the round is completable and we've cast votes.
+			// this is a workaround for avoiding restarting the round based on
```
gnunicorn: "based on " ... I guess we are missing something here?

src/voter.rs
rphmeier: no, iff is shorthand for "if and only if"

Issue #19, opened 2018-10-26
COMMIT messages and skipping ahead rounds


rphmeier:
A safe protocol for skipping ahead in rounds in a `voter` would be useful.  Perhaps it can be done outside, but with a utility provided by this crate.

COMMIT messages need to be fleshed out for observers to more easily witness finality.

rphmeier:
About skipping ahead in rounds: 

The voter keeps an `active_round: VotingRound` and a `prospective_skip: Option<VotingRound>`;

Whenever we see a commit for round `r > active_round.round + 1`, we start `prospective_skip` at `r+1` with no `last_round_estimate`.

Then one of three things will happen:
  - We see an even further commit and overwrite `prospective_skip`
  - `active_round` works its way up to `r` and becomes completable. In this case, we background `active_round`, set `active_round = prospective_skip.take()` and set the new `active_round.last_round_estimate` accordingly.
  - `prospective_skip` becomes completable, and we background it and start `active_round` at `r+2`.

Issue #9, opened 2018-09-10
Add Primary


rphmeier:

rphmeier:
The primary, on round-start, broadcasts the last round's estimate if they have not finalized it.
Voters alter their voting rules based on this block, when they've received it, but this is implemented already.

marcio-diaz:
Closed by #54

PR #25, opened 2018-11-14
Implement commit protocol

andresilva:
Related #9 #19.

- When a round is completable and we precommitted we start a commit timer and start accepting commit messages;
- When we receive a commit message if it targets a block higher than what we've finalized we validate it and import its precommits if valid;
- When our commit timer triggers we check if we've received any commit message for a block equal to what we've finalized, if we haven't then we broadcast a commit.

Also fixed some unrelated issues with the existing tests.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -201,6 +201,7 @@ impl<H, N: Copy, S, Id> SignedMessage<H, N, S, Id> {
 	}
 }
 
+/// A commit message which is an aggregate of precommits.
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "derive-codec", derive(Encode, Decode))]
 pub struct Commit<H, N, S, Id> {
@@ -212,6 +213,7 @@ pub struct Commit<H, N, S, Id> {
 	pub precommits: Vec<SignedPrecommit<H, N, S, Id>>,
 }
 
+/// A signed precommit message.
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "derive-codec", derive(Encode, Decode))]
 pub struct SignedPrecommit<H, N, S, Id> {
@@ -223,6 +225,7 @@ pub struct SignedPrecommit<H, N, S, Id> {
 	pub id: Id,
 }
 
+/// A commit message with compact representation of authentication data.
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "derive-codec", derive(Encode, Decode))]
 pub struct CompactCommit<H, N, S, Id> {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -264,6 +264,11 @@ impl<H: Clone, N: Clone, S, Id> From<Commit<H, N, S, Id>> for CompactCommit<H, N
 	}
 }
 
+fn threshold(total_weight: u64) -> u64 {
+	let faulty = total_weight.saturating_sub(1) / 3;
+	total_weight - faulty
+}
+
 #[cfg(test)]
 mod tests {
 	#[cfg(feature = "derive-codec")]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -223,17 +223,6 @@ pub struct SignedPrecommit<H, N, S, Id> {
 	pub id: Id,
 }
 
-#[derive(Debug, Clone, PartialEq, Eq)]
-#[cfg_attr(feature = "derive-codec", derive(Encode, Decode))]
-pub struct SignedCommit<H, N, S, Id> {
-	/// The commit message which has been signed.
-	pub commit: Commit<H, N, S, Id>,
-	/// The signature on the message.
-	pub signature: S,
-	/// The Id of the signer.
-	pub id: Id,
-}
-
 #[cfg(test)]
 mod tests {
 	#[cfg(feature = "derive-codec")]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -201,6 +201,39 @@ impl<H, N: Copy, S, Id> SignedMessage<H, N, S, Id> {
 	}
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "derive-codec", derive(Encode, Decode))]
+pub struct Commit<H, N, S, Id> {
+	/// The target block's hash.
+	pub target_hash: H,
+	/// The target block's number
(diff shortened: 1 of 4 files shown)


src/round.rs
```diff
@@ -166,10 +184,19 @@ impl<Id: Hash + Eq + Clone, Vote: Clone + Eq, Signature: Clone> VoteTracker<Id,
...
-				&*occupied.into_mut()
+				Some(&*occupied.into_mut())
 			}
 		}
 	}
+
+	fn valid_votes<'a>(&'a self) -> impl Iterator<Item=(Id, Vote, Signature)> + 'a {
+		self.votes.iter().filter_map(|(id, vote)| {
+			match vote {
+				VoteMultiplicity::Single(v, s) => Some((id.clone(), v.clone(), s.clone())),
```
rphmeier: I'm not sure what the repercussions of ignoring equivocations will be here. We may end up in situations where we have finalized blocks with equivocations, but cannot craft a commit message.

src/lib.rs
```diff
@@ -201,6 +201,39 @@ impl<H, N: Copy, S, Id> SignedMessage<H, N, S, Id> {
...
+	pub precommit: Precommit<H, N>,
+	/// The signature on the message.
+	pub signature: S,
+	/// The Id of the signer.
+	pub id: Id,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "derive-codec", derive(Encode, Decode))]
+pub struct SignedCommit<H, N, S, Id> {
```
rphmeier: AFAIK commit messages don't need to be signed again

src/voter.rs
```diff
@@ -441,16 +465,203 @@ impl<H, N, E: Environment<H, N>> Future for BackgroundRound<H, N, E> where
...
+		// ignore commits for a lower block than we already finalized
+		if commit.target_number < voting_round.votes.finalized().map(|(_, n)| *n).unwrap_or(N::zero()) {
+			return Ok(true);
+		}
+
+		// check that all precommits are for blocks higher than the target
+		// commit block, and that they're its descendents
+		if !commit.justification.iter().all(|signed| {
+			signed.precommit.target_number >= commit.target_number &&
+				env.ancestry(
```
rphmeier: precommits could also be for the block itself, in which case it would be `target_hash == commit.target_hash || ancestry(...).is_ok()`

src/voter.rs
```diff
@@ -441,16 +465,203 @@ impl<H, N, E: Environment<H, N>> Future for BackgroundRound<H, N, E> where
...
+
+impl<H, N, E: Environment<H, N>> RoundCommitter<H, N, E> where
+	H: Hash + Clone + Eq + Ord + ::std::fmt::Debug,
+	N: Copy + BlockNumberOps + ::std::fmt::Debug,
+{
+	fn import_commit(
+		&mut self,
+		env: &E,
+		commit: Commit<H, N, E::Signature, E::Id>,
+	) -> Result<bool, E::Error> {
```
rphmeier: i'm not sure if there is a check in this function against equivocations.

Is it possible to reuse the vote-graph and round logic for these checks somehow?

src/voter.rs
```diff
@@ -441,16 +465,203 @@ impl<H, N, E: Environment<H, N>> Future for BackgroundRound<H, N, E> where
...
+
+		// import all precommits
+		for SignedPrecommit { precommit, signature, id } in commit.justification.clone() {
+			if let Some(e) = voting_round.votes.import_precommit(env, precommit, id, signature)? {
+				env.precommit_equivocation(voting_round.votes.number(), e);
+			}
+		}
+
+		// TODO: is this right?
+		assert_eq!(
```
rphmeier: probably not right, because we could have finalized something even higher with the new precommits.

src/voter.rs
```diff
@@ -441,16 +465,203 @@ impl<H, N, E: Environment<H, N>> Future for BackgroundRound<H, N, E> where
...
+					commit.target_hash.clone(),
+					signed.precommit.target_hash.clone(),
+				).is_ok()
+		}) {
+			return Ok(false);
+		}
+
+		// check that the total vote weight of all precommits is higher than the threshold
+		let commit_weight = commit.justification.iter().fold(0, |total_weight, signed| {
+			total_weight + voting_round.votes.weight(&signed.id).unwrap_or(0)
```
rphmeier: this needs to discount equivocation weight

andresilva:
I think this is ready for another review.

src/lib.rs
```diff
@@ -201,6 +201,77 @@ impl<H, N: Copy, S, Id> SignedMessage<H, N, S, Id> {
...
+	pub target_number: N,
+	/// Precommits for target block or any block after it that justify this commit.
+	pub precommits: Vec<Precommit<H, N>>,
+	/// Authentication data for the commit.
+	pub auth_data: CommitAuthData<S, Id>,
+}
+
+// Authentication data for a commit, currently a set of precommit signatures but
+// in the future could be optimized with BLS signature aggregation.
+type CommitAuthData<S, Id> = Vec<(S, Id)>;
```
rphmeier: needs to be a pub type, yeah?

src/voter.rs
```diff
@@ -441,16 +480,214 @@ impl<H, N, E: Environment<H, N>> Future for BackgroundRound<H, N, E> where
...
+		try_ready!(self.commit_timer.poll());
+
+		let voting_round = self.voting_round.lock();
+		let commit = || -> Option<Commit<H, N, E::Signature, E::Id>> {
+			let (target_hash, target_number) = voting_round.votes.finalized().cloned()?;
+
+			let mut ids = HashSet::new();
+			let precommits =
+				voting_round.votes.precommits().into_iter().filter_map(|(id, precommit, signature)| {
+					match env.ancestry(target_hash.clone(), precommit.target_hash.clone()) {
```
rphmeier: this will exclude direct votes on the block I think.
since this is a common pattern we might want to add a new function on `Chain` with provided impl: 
```rust
fn is_equal_or_descendent_of(&self, base: H, block: H) -> Result<bool> {
    if base == block { return Ok(true) }
    
    match self.ancestry(base, block) {
         Ok(_) => Ok(true),
         Err(NotDescendent) => Ok(false),
         Err(e) => Err(e),
    }
}
```

andresilva: I think it works with the current implementation in substrate that uses `tree_route`, but I agree it's better to add that method anyway.

PR #28, opened 2018-11-20
Round skipping

### LINKED ISSUE #19: COMMIT messages and skipping ahead rounds

andresilva:
Closes #19.

~Still needs to be cleaned up before merging.~

### DIFF
--- a/src/voter.rs
+++ b/src/voter.rs
@@ -846,6 +846,11 @@ impl<H, N, E: Environment<H, N>> Voter<H, N, E> where
 
 			// we saw a commit for a round `r` that is at least 2 higher than
 			// our current best round so we start a prospective round at `r + 1`
+			//
+			// we also start a prospective round only if our last prospective round is before
+			// the given commit message. we could technically restart if they are the same,
+			// but if commit messages at the round are live then other messages are likely to be
+			// as well. Not restarting gives the best chance of completing the round faster.
 			let should_start_prospective = round_number > self.best_round.votes.number() + 1 &&
 				prospective_round_number.map_or(true, |n| round_number > n);
 
@@ -863,7 +868,6 @@ impl<H, N, E: Environment<H, N>> Voter<H, N, E> where
 					//   - a finalized block in the active round, or
 					//   - the last finalized in prior rounds
 					let ghost_base = self.prospective_round.as_ref()
-						.filter(|r| r.votes.number() + 1 < round_number)
 						.and_then(|r| r.votes.state().finalized.clone())
 						.or_else(|| self.best_round.votes.state().finalized.clone())
 						.unwrap_or_else(|| self.last_finalized_in_rounds.clone());
--- a/src/voter.rs
+++ b/src/voter.rs
@@ -863,6 +863,7 @@ impl<H, N, E: Environment<H, N>> Voter<H, N, E> where
 					//   - a finalized block in the active round, or
 					//   - the last finalized in prior rounds
 					let ghost_base = self.prospective_round.as_ref()
+						.filter(|r| r.votes.number() + 1 < round_number)
 						.and_then(|r| r.votes.state().finalized.clone())
 						.or_else(|| self.best_round.votes.state().finalized.clone())
 						.unwrap_or_else(|| self.last_finalized_in_rounds.clone());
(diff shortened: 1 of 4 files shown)


src/voter.rs
```diff
@@ -803,63 +831,132 @@ impl<H, N, E: Environment<H, N>> Voter<H, N, E> where
...
-		let mut last_finalized_number = self.last_finalized_number.lock();
-		if finalized_number > *last_finalized_number {
-			*last_finalized_number = finalized_number;
-			return true;
+	fn process_commits(&mut self) -> Result<(), E::Error> {
+		if let Async::Ready((round_number, commit)) = self.committer.poll()? {
+			let prospective_round_number =
+				self.prospective_round.as_ref().map(|r| r.votes.number());
+
+			// we saw a commit for a round `r` that is higher than our current best
```
rphmeier: comment should clarify: at least 2 higher

src/voter.rs
```diff
@@ -803,63 +831,132 @@ impl<H, N, E: Environment<H, N>> Voter<H, N, E> where
...
+			// round so we start a prospective round at `r + 1`
+			if round_number > self.best_round.votes.number() + 1 &&
+				prospective_round_number.iter().all(|n| round_number > *n) {
```
rphmeier: this is an `Option`, right? `map_or` would be more clear. Right now it's not 100% obvious what happens with `all` when it's `None`.

src/voter.rs
```diff
@@ -803,63 +831,132 @@ impl<H, N, E: Environment<H, N>> Voter<H, N, E> where
...
+					// then we background the current `best_round` and set the `best_round`
+					// to `prospective_round`.
+					if self.best_round.votes.number() == prospective_round.votes.number() - 1 {
+						trace!(target: "afg", "Best round at {} has caught up with prospective round at {}. \
+											   Setting best round to prospective round.",
+							   self.best_round.votes.number(),
+							   prospective_round.votes.number());
+
+						prospective_round.last_round_state = Some(self.best_round.bridge_state());
+						self.completed_best_round(Some(prospective_round))?;
```
rphmeier: the `prospective_round` isn't completed here, is it? The `best_round` is though. Also it doesn't look like the get swapped anymore

rphmeier: ahh, I misread; I see now that `completed_best_round` takes the next round.

src/voter.rs
```diff
@@ -869,10 +966,64 @@ impl<H, N, E: Environment<H, N>> Future for Voter<H, N, E> where
...
+
+	fn completed_prospective_round(&mut self, mut prospective_round: VotingRound<H, N, E>)
+		-> Result<(), E::Error>
+	{
+		self.env.completed(prospective_round.votes.number(), prospective_round.votes.state())?;
+
+		self.best_round = VotingRound::new(
+			prospective_round.votes.number() + 1,
+			// FIXME: this may be behind, the `prospective_round` won't issue
+			// finalization notifications (since it doesn't vote)
```
rphmeier: i'll ask alistair what to do here; it's possible we could take `prospective_round.votes.state().finalized` if it's higher than `last_finalized_in_rounds`

andresilva: Or we could at least use the finalized commit target that triggered the prospective round.

rphmeier: (it's also not the worst thing if this is behind but it also could be a bit of a performance hit if we have to construct large DAGs)

rphmeier: let's leave as is for now. we should just optimize the `vote_graph` and `ancestry` if this ends up being too slow in practice

rphmeier: > Or we could at least use the finalized commit target that triggered the prospective round.

this sounds right

src/voter.rs
```diff
@@ -272,6 +272,14 @@ impl<H, N, E: Environment<H, N>> VotingRound<H, N, E> where
 			trace!(target: "afg", "Got incoming message");
 			let SignedMessage { message, signature, id } = incoming;
 
+			if !self.env.is_equal_or_descendent_of(self.votes.base().0, message.target().0.clone()) {
+				trace!(target: "afg", "Ignoring message targeting {:?} lower than round base {:?}",
```
rphmeier: in the future, I'm pretty sure we could actually report misbehavior on this if we had a stored commit from this voter from recently.

Issue #30, opened 2018-11-21
Release 0.4


rphmeier:
Remaining refactors:
  - Remove `voters` method from `Environment`
  - Have `Voter` take commit message communication stream/sink on initialization

PR #31, opened 2018-11-22
Minor API changes for grandpa 0.4.0

andresilva:
Removed `voters` and the commit streams out of `Environment`, these are passed to the `Voter` instead when creating it since they're static through its lifetime. I think I actually prefer having the commit streams declared in the `Environment` since it leads to less type pollution in the definition of the `Committer` and `Voter`.

Ref #30

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "finality-grandpa"
-version = "0.3.0"
+version = "0.4.0"
 description = "PBFT-based finality gadget for blockchains"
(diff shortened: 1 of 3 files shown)


rphmeier:
We could still define the types of `CommitIn` and `CommitOut` in the environment.

It's nicer IMO not to have to define those types at all, especially because it means that in practice we probably have to box them.

andresilva:
Don't merge this yet I still want to make some changes and update the substrate PR with these changes (https://github.com/paritytech/substrate/pull/1151).

PR #33, opened 2018-12-04
Expose commit validation

andresilva:
Used for validating justifications in https://github.com/paritytech/substrate/pull/1190

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "finality-grandpa"
-version = "0.5.0"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "finality-grandpa"
-version = "0.5.0"
+version = "0.6.0"
```
rphmeier: this is a backwards-compatible change, right? so we could have version = `0.5.1`

PR #36, opened 2018-12-19
count equivocations on all votes

### LINKED ISSUE #34: Count equivocations as voting for everything
Currently, equivocations are counted on the blocks they target.

If we count equivocations as equivocating for everything, then we can safely ignore all future votes from an authority after seeing one equivocation.

rphmeier:
Closes #34 

The previous behavior was that equivocations would be counted on all branches that were actually equivocated for. Now, if we see a voter double-voting, we count it as "infinite-voting" and treat it as if we have received a vote for every possible block. This means that we can ignore any equivocations beyond the first.

Rewrites most of the `bitfield` module. The idea is that now, each `VoteWeight` is its own bitfield which counts all the voters who have voted on a specific node in the graph, and there is also a round-global bitfield of all equivocators.

When calculating the vote-weights corresponding to a node, we calculate the union of `node_bitfield` and `equivocators_bitfield`, and figure out the total weight based on that. I couldn't figure out a more allocation-efficient way of doing this. A pool for bitfield storage might be an improvement.

Since we now have a need for a global-ish lookup of validator indices to weights (before it was only equivocators -> weights), I introduced the `VoterSet` type which is used in lieu of `HashMap<Id, u64>` in the codebase.

### DIFF
--- a/src/round.rs
+++ b/src/round.rs
@@ -80,7 +80,7 @@ impl AddAssign for VoteWeight {
 enum VoteMultiplicity<Vote, Signature> {
 	// validator voted once.
 	Single(Vote, Signature),
-	// validator equivocated once.
+ 	// validator equivocated at least once.
 	Equivocated((Vote, Signature), (Vote, Signature)),
 }
 
--- a/src/round.rs
+++ b/src/round.rs
@@ -22,44 +22,57 @@ use std::collections::hash_map::{HashMap, Entry};
 use std::hash::Hash;
 use std::ops::AddAssign;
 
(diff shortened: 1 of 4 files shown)


src/round.rs
```diff
@@ -69,59 +82,9 @@ enum VoteMultiplicity<Vote, Signature> {
 	Single(Vote, Signature),
 	// validator equivocated once.
```
andresilva:
```suggestion
 	// validator equivocated at least once.
```

PR #38, opened 2018-12-21
adjust base in vote_graph

rphmeier:
This will be a helpful utility for doing accountable safety proofs; basically when we are answering queries we don't know what the base should be. 

The method will be to start with the block we try to prove impossibility of estimate, and then import historical votes into a `round` until we get `NotDescendent`, and then compute the common ancestor, adjust the base, and continue importing.

Might be useful for `commit` validation as well.

### DIFF
--- a/src/vote_graph.rs
+++ b/src/vote_graph.rs
@@ -119,6 +119,49 @@ impl<H, N, V> VoteGraph<H, N, V> where
 		(self.base.clone(), self.base_number)
 	}
 
+	/// Adjust the base of the graph. The new base must be an ancestor of the
+	/// old base.
+	///
+	/// Provide an ancestry proof from the old base to the new. The proof
+	/// should be in reverse order from the old base's parent.
+	pub fn adjust_base(&mut self, ancestry_proof: &[H]) {
+		let new_hash = match ancestry_proof.last() {
+			None => return, // empty -- nothing to do.
+			Some(h) => h,
+		};
+
+		// not a valid ancestry proof. TODO: error?
+		if ancestry_proof.len() > self.base_number.as_() { return }
+
+		// hack because we can't convert usize -> N, only vice-versa.
+		// hopefully LLVM can optimize.
+		let new_number = {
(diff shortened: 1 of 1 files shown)


src/vote_graph.rs
```diff
@@ -119,6 +119,49 @@ impl<H, N, V> VoteGraph<H, N, V> where
...
+		let new_hash = match ancestry_proof.last() {
+			None => return, // empty -- nothing to do.
+			Some(h) => h,
+		};
+
+		// not a valid ancestry proof. TODO: error?
+		if ancestry_proof.len() > self.base_number.as_() { return }
+
+		// hack because we can't convert usize -> N, only vice-versa.
+		// hopefully LLVM can optimize.
```
andresilva: Maybe we could add `From<u64>` to `BlockNumberOps`.

src/vote_graph.rs
```diff
@@ -119,6 +119,49 @@ impl<H, N, V> VoteGraph<H, N, V> where
...
+	/// Provide an ancestry proof from the old base to the new. The proof
+	/// should be in reverse order from the old base's parent.
+	pub fn adjust_base(&mut self, ancestry_proof: &[H]) {
+		let new_hash = match ancestry_proof.last() {
+			None => return, // empty -- nothing to do.
+			Some(h) => h,
+		};
+
+		// not a valid ancestry proof. TODO: error?
+		if ancestry_proof.len() > self.base_number.as_() { return }
```
andresilva: Maybe this should be `>=`?

src/vote_graph.rs
rphmeier: I wanted to avoid that because it would probably cause issues with `u32` block numbers (very likely in ephemeral chains).  A `TryFrom` implementation would  be fine. i'll add a TODO.

src/vote_graph.rs
rphmeier: I think we support block `0` so block `1` could have ancestry of `1`.

PR #39, opened 2019-01-04
Validate commits with Round and generate commits under new equivocation rules

### LINKED ISSUE #37: Use Round when validating commits
Currently `validate_commit` uses the `vote_graph` directly as a workaround for the fact that there is no `precommit_ghost` in `round`. This should be added and then `round` should be used for checking commits so that equivocations can be handled correctly.

rphmeier:
Closes #37 

We now generate commits with any votes targeting the chain or any equivocations (as they count for everything).

We could be a bit more clear about the preconditions/param ownership of `validate_commit` but that's a bit beyond the scope here.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -389,7 +389,7 @@ pub fn validate_commit<H, N, S, I, C: Chain<H, N>>(
 	let mut equivocated = std::collections::HashSet::new();
 
 	// Add all precommits to the round with correct counting logic
-	// the commit target as a base.
+	// using the commit target as a base.
 	let mut round = round::Round::new(round::RoundParams {
 		round_number: 0, // doesn't matter here.
 		voters: voters.clone(),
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -382,45 +386,26 @@ pub fn validate_commit<H, N, S, I, C: Chain<H, N>>(
...
-		return Ok(None);
-	}
+	let mut equivocated = std::collections::HashSet::new();
 
-	// make sure weight of all precommits surpasses threshold
-	// (this is needed to avoid a possible DoS vector)
-	let commit_weight = commit.precommits.iter().fold(0, |total_weight, precommit| {
-		total_weight + voters.info(&precommit.id).map(|info| info.weight()).unwrap_or(0)
+	// Add all precommits to the round with correct counting logic
+	// the commit target as a base.
```
andresilva:
```suggestion
	// using the commit target as a base.
```

PR #41, opened 2019-01-24
update codec to 0.3

thiolliere:
this is a breaking change as codec is exposed.
Then we can release a new version so we can use new codec in substrate


thiolliere:
also release a new major. I'm not sure about the policy (does new release be in its own PR ?) but the process seems fine as is.

thiolliere:
Ok so revert the commit to only update codec to 0.3

gnunicorn:
[clabot:check]

ddorgan:
[clabot:check]

Issue #48, opened 2019-02-24
Split voter module into multiple files


rphmeier:
Probably:
  - voting_round
  - background_rounds
  - voter
  - tests

Issue #47, opened 2019-02-18
Keep messages from recent rounds


rphmeier:
Related to #27 

Normally, when catching up, if we see a commit message for round R, we then start a prospective round at R+1 and upon seeing that complete, we start voting at R+2.

A catch-up message is all the messages for a given round. Importing this enables a voter to start voting in the following round. This is a bit more robust than the prospective round strategy that was previously implemented.

andresilva:
Same as #27, the requesting side is implemented in Substrate. We also added new `Environment` hooks that allow persisting all votes until a round is concluded. Do you think we can close this one as well?

rphmeier:
Yeah, we can close this.

PR #49, opened 2019-02-24
Refactors voter to support multiple kinds of global messages

rphmeier:
(the global messages are added in this PR for #47 but are currently ignored).

This also removes the shared ownership and mutexes that was used in the `Committer` architecture.
Will be followed up by #48.


src/voter/mod.rs
```diff
@@ -849,49 +947,109 @@ impl<H, N, E: Environment<H, N>, CommitIn, CommitOut> Voter<H, N, E, CommitIn, C
...
+					// that returns Some if there wasn't one.
+					if let Some(commit) = self.past_rounds.import_commit(round_number, commit) {
+						// otherwise validate the commit and signal the finalized block
+						// (if any) to the environment
+						if let Some((finalized_hash, finalized_number)) = validate_commit(
+							&commit,
+							&self.voters,
+							&*self.env,
+						)? {
+							match highest_incoming_foreign_commit {
```
gnunicorn: Is this equal to 
```rust
highest_incoming_foreign_commit = highest_incoming_foreign_commit.map_or(round_number,
	|n| ::std::cmp::max(round_number, n));
```
?

Then I'd prefer this, because I feel it is more expressive of what is happening than the current match.

src/voter/mod.rs
```diff
@@ -849,49 +947,109 @@ impl<H, N, E: Environment<H, N>, CommitIn, CommitOut> Voter<H, N, E, CommitIn, C
...
+		// we also start a prospective round only if our last prospective round is before
+		// the given commit message. we could technically restart if they are the same,
+		// but if commit messages at the round are live then other messages are likely to be
+		// as well. Not restarting gives the best chance of completing the round faster.
+		let should_start_prospective = round_number > self.best_round.votes.number() + 1 &&
+			prospective_round_number.map_or(true, |n| round_number > n);
+
+		if should_start_prospective {
+			trace!(target: "afg", "Imported commit for later round than current best {}, starting prospective round at {}",
+				   self.best_round.votes.number(),
```
gnunicorn: spaces?
```suggestion
				self.best_round.votes.number(),
```

src/voter/mod.rs
```diff
@@ -849,49 +947,109 @@ impl<H, N, E: Environment<H, N>, CommitIn, CommitOut> Voter<H, N, E, CommitIn, C
...
+				   round_number + 1);
```
gnunicorn: spaces and `);` always ends on the ident level of opening `(`...
```suggestion
				round_number + 1
			);
```

src/voter/mod.rs
```diff
@@ -1040,20 +1190,14 @@ impl<H, N, E: Environment<H, N>, CommitIn, CommitOut> Voter<H, N, E, CommitIn, C
 			self.env.clone(),
 		);
 
-		let background = BackgroundRound {
-			inner: Arc::new(Mutex::new(prospective_round)),
-			task: None,
-			finalized_number: N::zero(), // TODO: do that right.
-		};
-
 		// TODO: should we clear `past_rounds`?
```
gnunicorn: for sure need to at some point, no? Just for memory usage reasons...

Maybe put this into a ticket?

src/voter/mod.rs
```diff
@@ -480,48 +537,76 @@ struct BackgroundRound<H, N, E: Environment<H, N>> where
...
 }
 
+enum BackgroundRoundChange<H, N, E: Environment<H, N>> where
+	H: Hash + Clone + Eq + Ord + ::std::fmt::Debug,
+	N: Copy + BlockNumberOps + ::std::fmt::Debug,
+{
+	/// Background round has become irrelevant and can be discarded.
+	Irrelevant(u64),
+	/// Background round has a commit message to issue but should continue
+	/// being drive afterwards.
```
andresilva:
```suggestion
	/// being driven afterwards.
```

src/voter/mod.rs
```diff
@@ -591,179 +694,181 @@ impl<H, N, E: Environment<H, N>> RoundCommitter<H, N, E> where
...
-					// (due to the call to `self.rounds.get_mut`).
-					let mut last_finalized_number = self.last_finalized_number.lock();
+	// push an old voting round onto this stream.
+	fn push(&mut self, env: &E, round: VotingRound<H, N, E>) {
+		let round_number = round.votes.number();
+		let (tx, rx) = mpsc::unbounded();
+		let background = BackgroundRound {
+			inner: round,
+			task: None,
+			finalized_number: N::zero(), // TODO: do that right.
```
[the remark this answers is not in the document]
rphmeier: it's an old TODO, not from this PR unless you count moving code around.

src/voter/mod.rs
rphmeier: `past_rounds` is already slowly cleared as new blocks are finalized and old rounds become irrelevant

rphmeier:
Grumbles addressed.

PR #52, opened 2019-03-11
Fs ghost symmetry test

montekki:
it looks like while searching for GHOST we always take the first descendant that fits the constraint, while we should be taking the highest.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -182,17 +182,13 @@ pub trait Chain<H: Eq, N: Copy + BlockNumberOps> {
 	fn best_chain_containing(&self, base: H) -> Option<(H, N)>;
 
 	/// Returns true if `block` is a descendent of or equal to the given `base`.
-	fn is_equal_or_descendent_of(&self, base: H, block: H) -> bool {
-		if base == block { return true; }
+	fn is_equal_or_descendent_of(&self, base: H, block: H) -> Result<bool, Error> {
+		if base == block { return Ok(true); }
 
-		// TODO: currently this function always succeeds since the only error
-		// variant is `Error::NotDescendent`, this may change in the future as
-		// other errors (e.g. IO) are not being exposed.
 		match self.ancestry(base, block) {
-			Ok(_) => true,
-			Err(Error::NotDescendent) => false,
-			// TODO: probably this whole error handling match should be re-thinked now
-			Err(Error::MultipleForksSupportPredicate) => unreachable!(),
+			Ok(_) => Ok(true),
+			Err(Error::NotDescendent) => Ok(false),
+			Err(e) => Err(e) ,
 		}
 	}
 }
@@ -415,7 +411,7 @@ pub fn validate_commit<H, N, S, I, C: Chain<H, N>>(
 			chain.is_equal_or_descendent_of(
 				commit.target_hash.clone(),
 				signed.precommit.target_hash.clone(),
-			)
+			).unwrap_or(false)
 	}) {
 		return Ok(None);
 	}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -406,14 +406,14 @@ pub fn validate_commit<H, N, S, I, C: Chain<H, N>>(
 {
 	// check that all precommits are for blocks higher than the target
 	// commit block, and that they're its descendents
-	if !commit.precommits.iter().all(|signed| {
-		signed.precommit.target_number >= commit.target_number &&
+	for signed in commit.precommits.iter() {
+		if !(signed.precommit.target_number >= commit.target_number &&
 			chain.is_equal_or_descendent_of(
 				commit.target_hash.clone(),
 				signed.precommit.target_hash.clone(),
-			).unwrap_or(false)
-	}) {
-		return Ok(None);
+			)?) {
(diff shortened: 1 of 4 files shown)


src/vote_graph.rs
```diff
@@ -304,7 +304,7 @@ impl<H, N, V> VoteGraph<H, N, V> where
 					}
 				})
 				.filter(|&(_, ref node)| condition(&node.cumulative_vote))
-				.next();
+				.max_by(|&(_, ref node1), &(_, ref node2)| node1.number.cmp(&node2.number));
```
rphmeier: As mentioned in a Riot discussion, the two statements are actually equivalent. Note this line in the docs above:

https://github.com/paritytech/finality-grandpa/blob/12482fcbf36873e8c787e0012e9d25f5b36ec4af/src/vote_graph.rs#L261-L263

src/lib.rs
```diff
@@ -188,6 +191,8 @@ pub trait Chain<H: Eq, N: Copy + BlockNumberOps> {
 		match self.ancestry(base, block) {
 			Ok(_) => true,
 			Err(Error::NotDescendent) => false,
+			// TODO: probably this whole error handling match should be re-thinked now
+			Err(Error::MultipleForksSupportPredicate) => unreachable!(),
```
rphmeier: yup, there's a TODO above about it. Let's just make this function return a result.

montekki: done, but you probably mean
```rust
Ok(_) => Ok(true),
Err(Error::NotDescendent) => Ok(false),
Err(a) => Err(a),
```

src/lib.rs
```diff
@@ -410,7 +411,7 @@ pub fn validate_commit<H, N, S, I, C: Chain<H, N>>(
 			chain.is_equal_or_descendent_of(
 				commit.target_hash.clone(),
 				signed.precommit.target_hash.clone(),
-			)
+			).unwrap_or(false)
```
rphmeier: @andresilva i always forget what the semantics of `Ok(None)` are for `validate_commit`. That means that it validated successfully, but the result was that the commit is bad?

andresilva: Yeah, this function should return the precommit ghost, which should exist if the commit is well-formed and hence the function returns `Ok(Some(_))` when given a valid commit. I think in this case we shouldn't call `unwrap_or(false)` and instead just do `?`.

src/round.rs
```diff
@@ -451,7 +451,10 @@ impl<Id, H, N, Signature> Round<Id, H, N, Signature> where
 				if let VoteMultiplicity::Single(ref v, _) = *multiplicity {
 					// if there is a single vote from this voter, we only include it
 					// if it branches off of the target.
-					chain.is_equal_or_descendent_of(f_hash.clone(), v.target_hash.clone())
+					match chain.is_equal_or_descendent_of(f_hash.clone(), v.target_hash.clone()) {
+						Ok(b) => b,
+						Err(_) => false,
```
andresilva: I don't think this is correct, I think we should just return the error here. Although in this case it's a bit murky because we're only detecting this error when iterating.

src/vote_graph.rs
```diff
@@ -303,8 +303,12 @@ impl<H, N, V> VoteGraph<H, N, V> where
 						true
 					}
 				})
-				.filter(|&(_, ref node)| condition(&node.cumulative_vote))
-				.next();
+				.filter(|&(_, ref node)| condition(&node.cumulative_vote));
+			let next_descendent = descendent_candidates.next();
```
andresilva:
```suggestion
			
			let next_descendent = descendent_candidates.next();
```

src/voter/voting_round.rs
```diff
@@ -274,7 +274,7 @@ impl<H, N, E: Environment<H, N>> VotingRound<H, N, E> where
 					// the ancestor of the current round's p-Ghost before precommitting.
 					self.votes.state().prevote_ghost.as_ref().map_or(false, |p_g| {
 						p_g == &last_round_estimate ||
-							self.env.is_equal_or_descendent_of(last_round_estimate.0, p_g.0.clone())
+							self.env.is_equal_or_descendent_of(last_round_estimate.0, p_g.0.clone()).unwrap_or(false)
```
andresilva:
```suggestion
							self.env.is_equal_or_descendent_of(last_round_estimate.0, p_g.0.clone())?
```

src/voter/voting_round.rs
```diff
@@ -343,6 +343,7 @@ impl<H, N, E: Environment<H, N>> VotingRound<H, N, E> where
 							}
 						}
 						Err(crate::Error::NotDescendent) => last_round_estimate.0,
+						Err(crate::Error::MultipleForksSupportPredicate) => { return Err(crate::Error::MultipleForksSupportPredicate.into()) },
```
andresilva:
```suggestion
						Err(e) => { return Err(e.into()); },
```

montekki: is this a good thing to do? when a new error will be added it would silently wildcard here and the exhausive pattern checking would be of no help

andresilva:
Closing this as stale, not sure it's worth it to add this check.

PR #53, opened 2019-03-14
replace prospective round machinery with catch-up rounds

rphmeier:
The existing process for catching up to the current round when behind followed these steps:
  - Watch for future commit messages
  - After a future commit, start a "prospective" round running in parallel to the best round we've currently got
  - If we see this prospective round conclude, we set our best round to be prospective + 1.

Replaces with catch-up rounds (#47). Logic for constructing and requesting catch-up rounds isn't yet implemented, but importing is done.

### DIFF
--- a/src/voter/mod.rs
+++ b/src/voter/mod.rs
@@ -514,6 +514,11 @@ impl<H, N, E: Environment<H, N>, GlobalIn, GlobalOut> Voter<H, N, E, GlobalIn, G
 
 						for prevote in &catch_up.prevotes {
 							if !voters.contains_key(&prevote.id) {
+								trace!(target: "afg",
+									"Ignoring invalid catch up, invalid voter: {:?}",
+									prevote.id,
+								);
+
 								process_catch_up_outcome.run(CatchUpProcessingOutcome::Bad(BadCatchUp::new()));
 								return Ok(())
 							}
@@ -522,6 +527,11 @@ impl<H, N, E: Environment<H, N>, GlobalIn, GlobalOut> Voter<H, N, E, GlobalIn, G
 
 						for precommit in &catch_up.precommits {
 							if !voters.contains_key(&precommit.id) {
+								trace!(target: "afg",
+									"Ignoring invalid catch up, invalid voter: {:?}",
+									precommit.id,
+								);
+
 								process_catch_up_outcome.run(CatchUpProcessingOutcome::Bad(BadCatchUp::new()));
 								return Ok(())
 							}
@@ -547,6 +557,10 @@ impl<H, N, E: Environment<H, N>, GlobalIn, GlobalOut> Voter<H, N, E, GlobalIn, G
 
 						let threshold = voters.threshold();
 						if pv < threshold || pc < threshold {
+							trace!(target: "afg",
+								"Ignoring invalid catch up, missing voter threshold"
+							);
+
 							process_catch_up_outcome.run(CatchUpProcessingOutcome::Bad(BadCatchUp::new()));
 							return Ok(())
 						}
@@ -563,6 +577,11 @@ impl<H, N, E: Environment<H, N>, GlobalIn, GlobalOut> Voter<H, N, E, GlobalIn, G
 						match round.import_prevote(&*self.env, prevote, id, signature) {
 							Ok(_) => {},
 							Err(e) => {
+								trace!(target: "afg",
+									"Ignoring invalid catch up, error importing prevote: {:?}",
+									e,
+								);
+
 								process_catch_up_outcome.run(CatchUpProcessingOutcome::Bad(BadCatchUp::new()));
 								return Ok(());
 							},
@@ -574,6 +593,11 @@ impl<H, N, E: Environment<H, N>, GlobalIn, GlobalOut> Voter<H, N, E, GlobalIn, G
 						match round.import_precommit(&*self.env, precommit, id, signature) {
 							Ok(_) => {},
 							Err(e) => {
+								trace!(target: "afg",
+									"Ignoring invalid catch up, error importing precommit: {:?}",
+									e,
+								);
+
 								process_catch_up_outcome.run(CatchUpProcessingOutcome::Bad(BadCatchUp::new()));
 								return Ok(());
 							},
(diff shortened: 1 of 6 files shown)


src/voter/mod.rs
```diff
@@ -387,139 +352,133 @@ impl<H, N, E: Environment<H, N>, GlobalIn, GlobalOut> Voter<H, N, E, GlobalIn, G
...
-		// as well. Not restarting gives the best chance of completing the round faster.
-		let should_start_prospective = round_number > self.best_round.round_number() + 1 &&
-			prospective_round_number.map_or(true, |n| round_number > n);
-
-		if should_start_prospective {
-			trace!(target: "afg", "Imported commit for later round than current best {}, starting prospective round at {}",
-				self.best_round.round_number(),
-				round_number + 1,
-			);
+					// check threshold support in prevotes and precommits.
```
marcio-diaz: Maybe extract into own function.

src/voter/mod.rs
```diff
@@ -387,139 +352,133 @@ impl<H, N, E: Environment<H, N>, GlobalIn, GlobalOut> Voter<H, N, E, GlobalIn, G
...
-				self.voters.clone(),
-				ghost_base,
-				None,
-				self.best_round.finalized_sender(),
-				self.env.clone(),
-			));
-		}
+						for prevote in &catch_up.prevotes {
+							if !voters.contains_key(&prevote.id) {
+								return Ok(())
```
marcio-diaz: Maybe add a trace as above indicating why we ignore.

src/voter/mod.rs
```diff
@@ -387,139 +352,133 @@ impl<H, N, E: Environment<H, N>, GlobalIn, GlobalOut> Voter<H, N, E, GlobalIn, G
...
+						round_number: catch_up.round_number,
+						voters: self.voters.clone(),
+						base: (catch_up.base_hash, catch_up.base_number),
+					});
+
+					// import prevotes first.
+					for crate::SignedPrevote { prevote, id, signature } in catch_up.prevotes {
+						match round.import_prevote(&*self.env, prevote, id, signature) {
+							Ok(_) => {},
+							Err(e) => return Ok(()),
```
marcio-diaz: Unused error here.

src/voter/mod.rs
```diff
@@ -387,139 +352,133 @@ impl<H, N, E: Environment<H, N>, GlobalIn, GlobalOut> Voter<H, N, E, GlobalIn, G
...
 
-					} else {
-						// otherwise we keep the current `prospective_round` running.
-						self.prospective_round = Some(prospective_round);
-						best_round_completable = Some(false);
+					// then precommits.
+					for crate::SignedPrecommit { precommit, id, signature } in catch_up.precommits {
+						match round.import_precommit(&*self.env, precommit, id, signature) {
+							Ok(_) => {},
+							Err(e) => return Ok(()),
```
marcio-diaz: Same.

src/voter/mod.rs
```diff
@@ -387,139 +352,133 @@ impl<H, N, E: Environment<H, N>, GlobalIn, GlobalOut> Voter<H, N, E, GlobalIn, G
...
+
+					// update last-finalized in rounds _after_ starting new round.
+					// otherwise the base could be too eagerly set forward.
+					if let Some((f_hash, f_num)) = state.finalized.clone() {
+						if f_num > self.last_finalized_in_rounds.1 {
+							self.last_finalized_in_rounds = (f_hash, f_num);
+						}
+					}
+
+					self.env.completed(just_completed.round_number(), state)?;
```
marcio-diaz: Replace `state` by `just_completed.round_state()` for consistency?

src/voter/past_rounds.rs
```diff
@@ -60,8 +60,13 @@ impl<H, N, E: Environment<H, N>> BackgroundRound<H, N, E> where
 
 	fn is_done(&self) -> bool {
 		// no need to listen on a round anymore once the estimate is finalized.
+		//
+		// we map `None` to true because
+		//   - rounds are not backgrounded when incomplete unless we've skipped forward
```
marcio-diaz: Maybe add an `assert` in the corresponding function.

rphmeier:
for v0.8

andresilva:
This has been integrated into substrate and should be ready for a final review now.

src/voter/mod.rs
```diff
@@ -501,139 +495,179 @@ impl<H, N, E: Environment<H, N>, GlobalIn, GlobalOut> Voter<H, N, E, GlobalIn, G
...
+
+					let state = round.state();
+					if !state.completable {
+						process_catch_up_outcome.run(CatchUpProcessingOutcome::Bad(BadCatchUp::new()));
+						return Ok(())
+					}
+
+					// beyond this point, we set this round to the past and
+					// start voting in the next round.
+					let mut just_completed = VotingRound::completed(
```
rphmeier: why do we need this to be backgrounded at all?

src/voter/mod.rs
```diff
@@ -501,139 +495,179 @@ impl<H, N, E: Environment<H, N>, GlobalIn, GlobalOut> Voter<H, N, E, GlobalIn, G
...
+						just_completed.round_state(),
+						just_completed.dag_base(),
+						just_completed.historical_votes(),
+					)?;
+
+					self.past_rounds.push(&*self.env, just_completed);
+
+					// stop voting in the best round before we background it, just in case it contradicts what
+					// we're doing now.
+					self.best_round.stop_voting();
```
rphmeier: If the catch-up is to the round r+1 where `self.best_round` is at `r`, then it might be dangerous to stop voting. I think any contradiction would have to arise from others misbehaving, right?

rphmeier:
Generally looks fine, but I'm worried about the `stop_voting` stuff breaking invariants. Since every round is basing off of an _over_ estimate of the prior, as long as we follow the voting rules I don't think we can ever contradict what a future round has done _legitimately_.

src/voter/mod.rs
andresilva: As discussed, we might still get some votes on this round so it doesn't really hurt to keep it around.

PR #54, opened 2019-03-19
Add Primary

marcio-diaz:
PR for issue #9

### DIFF
--- a/src/bitfield.rs
+++ b/src/bitfield.rs
@@ -274,6 +274,14 @@ mod tests {
 	use super::*;
 	use crate::VoterSet;
 
+	fn to_prevote(id: usize) -> usize {
+		id * 2
+	}
+
+	fn to_precommit(id: usize) -> usize {
+		id * 2 + 1
+	}
+
 	#[test]
 	fn merge_live() {
 		let mut a = Bitfield::Live(LiveBitfield::with_voters(10));
@@ -288,11 +296,11 @@ mod tests {
 			(2, 7),
 		].iter().cloned().collect();
 
-		a.set_bit(0, 10).unwrap(); // prevote 1
-		a.set_bit(3, 10).unwrap(); // precommit 2
+		a.set_bit(to_prevote(v.info(&1).unwrap().canon_idx()), 10).unwrap(); // prevote 1
+		a.set_bit(to_precommit(v.info(&2).unwrap().canon_idx()), 10).unwrap(); // precommit 2
 
-		b.set_bit(4, 10).unwrap(); // prevote 3
-		b.set_bit(5, 10).unwrap(); // precommit 3
+		b.set_bit(to_prevote(v.info(&3).unwrap().canon_idx()), 10).unwrap(); // prevote 3
+		b.set_bit(to_precommit(v.info(&3).unwrap().canon_idx()), 10).unwrap(); // precommit 3
 
 		let c = a.merge(&b).unwrap();
 		assert_eq!(c.total_weight(|i| v.weight_by_index(i).unwrap()), (14, 16));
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -121,6 +121,22 @@ impl<H, N> Precommit<H, N> {
 	}
 }
 
+/// A Primary proposed block.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "derive-codec", derive(Encode, Decode))]
+pub struct Primary<H, N> {
```
rphmeier: nit: rename to `PrimaryPropose`

src/lib.rs
```diff
@@ -121,6 +121,22 @@ impl<H, N> Precommit<H, N> {
 	}
 }
 
+/// A Primary proposed block.
```
rphmeier: maybe a doc comment mentioning that this is a broadcast of the last round's estimate.

src/voter/voting_round.rs
```diff
@@ -232,12 +233,45 @@ impl<H, N, E: Environment<H, N>> VotingRound<H, N, E> where
 						self.env.precommit_equivocation(self.votes.number(), e);
 					}
 				}
+				Message::Primary(primary) => {
+					let primary_index = self.votes.number() as usize % self.votes.voters().voters.len();
```
rphmeier: even though this has access to `voters` I would prefer some APIs rather than direct field access

src/voter/voting_round.rs
```diff
@@ -232,12 +233,45 @@ impl<H, N, E: Environment<H, N>> VotingRound<H, N, E> where
...
 		Ok(())
 	}
 
+	fn primary(&mut self, last_round_state: &RoundState<H, N>) -> Result<(), E::Error> {
+		match self.state {
+			Some(State::Start(_, _)) => {
+				let last_round_estimate = last_round_state.estimate.clone()
+					.expect("Rounds only started when prior round completable; qed");
+				let last_round_finalized = last_round_state.finalized.clone()
+					.expect("Rounds only started when prior round completable; qed"); // TODO: check this.
```
rphmeier: these `expect`s are pretty subtle. Maybe it's better to log and just not propose if this happens?

src/lib.rs
```diff
@@ -456,7 +387,7 @@ pub fn validate_commit<H, N, S, I, C: Chain<H, N>>(
 	Ok(round.precommit_ghost())
 }
 
-fn threshold(total_weight: u64) -> u64 {
+pub fn threshold(total_weight: u64) -> u64 {
```
rphmeier: docs on public

src/voter/voting_round.rs
```diff
@@ -232,12 +234,47 @@ impl<H, N, E: Environment<H, N>> VotingRound<H, N, E> where
...
+						self.primary_block = Some((primary.target_hash, primary.target_number));
+					}
+				}
 			};
 		}
 
 		Ok(())
 	}
 
+	fn primary(&mut self, last_round_state: &RoundState<H, N>) -> Result<(), E::Error> {
```
rphmeier: `primary_propose`

src/voter/mod.rs
```diff
@@ -279,6 +280,7 @@ impl<H, N, E: Environment<H, N>, GlobalIn, GlobalOut> Voter<H, N, E, GlobalIn, G
 		last_round_number: u64,
 		last_round_state: RoundState<H, N>,
 		last_finalized: (H, N),
+		voter_id: Option<E::Id>,
```
rphmeier: this is getting pretty big. maybe refactor into a `VoterParams` struct?

src/bitfield.rs
```diff
@@ -289,7 +289,7 @@ mod tests {
 		].iter().cloned().collect();
 
 		a.set_bit(0, 10).unwrap(); // prevote 1
-		a.set_bit(11, 10).unwrap(); // precommit 2
+		a.set_bit(3, 10).unwrap(); // precommit 2
```
rphmeier: a comment here saying the `VoterSet` is sorting internally might be a good idea.

Issue #55, opened 2019-03-21
Expose information about the `Voter` to outside


rphmeier:
Right now, the only interaction that `Voter` has with the outside is informing the `Environment` when it starts a new round, completes one, finalizes a block, or detects an equivocation.

We should add a new type

```rust
#[derive(Clone)]
struct VoterState {
    inner: Arc<parking_lot::Mutex<Inner>>,
}

impl VoterState {
    // information like proposer, round number, what we've done in the round, etc.
    pub fn active_rounds(&self) -> Vec<RoundData>;
}

impl Voter {
    pub fn voter_state(&self) -> VoterState;
}
```

The `inner` member of `VoterState` should be updated by the voter whenever a round is added or dropped.

rphmeier:
Also, we want to expose an API for passing a callback along with a global message that is called with various return codes based on how useful or valid the message was.

andresilva:
Fixed in #114.

PR #57, opened 2019-03-27
voter: only primary voter proposes and only once

andresilva:


rphmeier:
We discussed this in the original PR and I was against including the `VoterId` in the voter at all. We haven't done it for any other messages either

andresilva:
Sorry I missed that discussion. What if the local id is exposed from the `Environment`? I don't really want to spam consensus gossip with useless messages, and we either filter here or on substrate (which means the logic for primary election leaks there).

rphmeier:
Yeah, although it does seem like a somewhat debatable point. I think we either go all-the-way with `VoterId` in `finality-grandpa` or not at all. So if `Id` is `None` we don't send messages at all, for instance.

rphmeier:
@andresilva 

> we either filter here or on substrate (which means the logic for primary election leaks there).

The idea is that with https://github.com/paritytech/finality-grandpa/issues/55 we will just call `voter_state.round_data(round).primary == self.id` for filtering -- the logic for primary election doesn't need to leak.

andresilva:
I agree that doing the filtering on substrate is more inline with what we have now, especially with #55. I'll update this PR shortly to remove this.

andresilva:
After discussing this with @rphmeier we've decided to go "all-in" on using the `voter_id` and move the logic for filtering messages to this crate. I've updated the PR so that no votes are sent out if we're not a current voter in the voter set. The local voter id is given per-round through `Environment::round_data`. The protocol state machine still transitions as if the votes were cast, the only difference being that we don't push the messages to the sink. Commits are still pushed out and should be filtered on substrate side (although it may be beneficial in some cases for observers to broadcast commits).

PR #58, opened 2019-03-28
Expose round votes through Environment

andresilva:
This is useful to persist votes and re-use them (if appropriate) when restarting the voter.

Depends on #57.


src/round.rs
```diff
@@ -616,6 +616,11 @@ impl<Id, H, N, Signature> Round<Id, H, N, Signature> where
 		self.voters.voter_by_index(self.round_number as usize % self.voters.len())
 	}
 
+	/// Return all imported precommits.
```
rphmeier: wrong comment

PR #59, opened 2019-04-03
Minor fixes for 0.7.0

andresilva:
- `CommunicationOut` and `CommunicationIn` are no longer encodable/decodable. The commit in `CommunicationOut` takes a callback that is not serializable. In general, these messages won't need to be encoded anyway.
- Removed the `Clone` on callback, seemed dangerous and it's only needed for tests.
- Removed unused file.
- Make fields of `CommitValidationResult` public.


src/lib.rs
```diff
@@ -338,11 +338,11 @@ impl<H: Clone, N: Clone, S, Id> From<Commit<H, N, S, Id>> for CompactCommit<H, N
...
-	ghost: Option<(H, N)>,
-	num_precommits: usize,
-	num_duplicated_precommits: usize,
-	num_equivocations: usize,
-	num_invalid_voters: usize,
+	pub ghost: Option<(H, N)>,
+	pub num_precommits: usize,
+	pub num_duplicated_precommits: usize,
+	pub num_equivocations: usize,
+	pub num_invalid_voters: usize,
```
rphmeier: Prefer getters since then we can add more without breaking API

src/voter/mod.rs
```diff
@@ -197,14 +191,14 @@ impl<H, N> From<CommitValidationResult<H, N>> for BadCommit {
...
 /// Callback used to propagate the commit in case the outcome is good.
-#[cfg_attr(feature = "derive-codec", derive(Encode, Decode))]
 pub enum Callback {
 	/// Default value.
 	Blank,
 	/// Callback to execute given a commit processing outcome.
 	Work(Box<FnMut(CommitProcessingOutcome) + Send>),
 }
 
+#[cfg(test)]
```
marcio-diaz: Thanks.

PR #60, opened 2019-04-05
Grant access to voters

cmichi:


src/voter_set.rs
```diff
@@ -62,6 +62,11 @@ impl<Id: Hash + Eq> VoterSet<Id> {
 	pub fn total_weight(&self) -> u64 {
 		self.voters.iter().map(|&(_, weight)| weight).sum()
 	}
+
+	/// Get the voters.
+	pub fn voters(&self) -> &Vec<(Id, u64)> {
```
rphmeier: `&Vec<>` is bad style. Should be `&[]`

rphmeier:
```suggestion
	pub fn voters(&self) -> &[(Id, u64)] {
```

Issue #10, opened 2018-09-10
Accountable Safety Protocol


rphmeier:
- Hook for dealing with safety violations.
- Issue and respond to challenges

Responding implies that we are able to return sets of votes such that we have some observable property. That probably means storing messages on disk in the order that we received them, and marking the points at which we voted. We will have to mark on the current and prior rounds, because one kind of query requires justification of the prior round's estimate at the point of voting.

rphmeier:
For now let's focus on the "answering queries" portion.

There are roughly two kinds of queries to answer (more details in paper).

1. What are all the prevotes you've seen for round r?

This is easy to answer as long as we keep all votes for that round.

2. Why was the estimate for round R !> B when you prevoted/precommited?

For this, we will want a way to mark which votes we had seen at the time when we prevoted and which we had seen at the time when we precommitted. To answer this query, we would simply query the DB and get the relevant set, and return it to it.


This implies that for each round r, a reasonable schema would be to store something like a 

```rust
struct HistoricalVotes {
    seen: Vec<Message>, // in order received
    prevote_idx: Option<usize>, // the length of `seen` at the time we prevoted, if we have.
    precommit_idx: Option<usize>, // the length of `seen` at the time we precommitted, if we have.
}
```

AlfonsoCev:
> For this, we will want a way to mark which votes we had seen at the time when we prevoted and which we had seen at the time when we precommitted.

This is not necessary. The answer to a challenge should be a set S of votes of the same type (all pre-votes or all pre-commits) and of the same round (for some round previous to the one referenced in the challenge) and with the property that "it is impossible for S to have a supermajority for B". But this property is preserved when we add new votes to S (see Lemma 2.2 (ii) in the Grandpa paper). So, if you saw such a set of votes at any point in the past, and you added some votes to that set since then, then the resulting set still has this property. 

Hence we just need to store all votes seen, for the last K rounds, for some large enough parameter K.

PR #65, opened 2019-05-08
Answer queries

marcio-diaz:
Starting #10 

This PR adds the `HistoricalVotes` struct, for saving votes in received order and indicating the moment of prevoting and precommiting.

### DIFF
--- a/src/round.rs
+++ b/src/round.rs
@@ -22,7 +22,7 @@ use crate::bitfield::{Shared as BitfieldContext, Bitfield};
 use crate::vote_graph::VoteGraph;
 use crate::voter_set::VoterSet;
 
-use super::{Equivocation, Prevote, Precommit, Chain, BlockNumberOps};
+use super::{Equivocation, Prevote, Precommit, Chain, BlockNumberOps, HistoricalVotes, Message};
 
 #[derive(Hash, Eq, PartialEq)]
 struct Address;
@@ -232,8 +232,7 @@ pub struct Round<Id: Hash + Eq, H: Hash + Eq, N, Signature> {
 	graph: VoteGraph<H, N, VoteWeight>, // DAG of blocks which have been voted on.
 	prevote: VoteTracker<Id, Prevote<H, N>, Signature>, // tracks prevotes that have been counted
 	precommit: VoteTracker<Id, Precommit<H, N>, Signature>, // tracks precommits
-	prevoted_indices: Option<(usize, usize)>,
-	precommited_indices: Option<(usize, usize)>,
+	historical_votes: HistoricalVotes<H, N>,
 	round_number: u64,
 	voters: VoterSet<Id>,
 	total_weight: u64,
@@ -285,8 +284,11 @@ impl<Id, H, N, Signature> Round<Id, H, N, Signature> where
 			graph: VoteGraph::new(base_hash, base_number),
 			prevote: VoteTracker::new(),
 			precommit: VoteTracker::new(),
-			prevoted_indices: None,
-			precommited_indices: None,
+			historical_votes: HistoricalVotes {
+				seen: Vec::new(),
+				prevote_idx: None,
+				precommit_idx: None,
+			},
 			bitfield_context: BitfieldContext::new(n_validators),
 			prevote_ghost: None,
 			precommit_ghost: None,
@@ -344,6 +346,8 @@ impl<Id, H, N, Signature> Round<Id, H, N, Signature> where
 						vote_weight,
 						chain,
 					)?;
+
+					self.historical_votes.seen.push(Message::Prevote(vote.clone()));
 
 					None
 				}
@@ -421,6 +425,8 @@ impl<Id, H, N, Signature> Round<Id, H, N, Signature> where
 						vote_weight,
 						chain,
 					)?;
+
+					self.historical_votes.seen.push(Message::Precommit(vote.clone()));
 
 					None
 				}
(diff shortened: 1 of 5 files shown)


src/round.rs
```diff
@@ -674,6 +676,30 @@ impl<Id, H, N, Signature> Round<Id, H, N, Signature> where
 	pub fn precommits(&self) -> Vec<(Id, Precommit<H, N>, Signature)> {
 		self.precommit.votes()
 	}
+
+	/// Set the length of prevotes received at the moment of prevoting.
+	pub fn set_prevote_idx(&mut self) -> bool {
+		self.prevote.voted_at = Some(self.prevote.votes().len());
+		println!("set prevote idx to {:?}", self.prevote.voted_at);
```
rphmeier: stray println

src/round.rs
```diff
@@ -674,6 +676,30 @@ impl<Id, H, N, Signature> Round<Id, H, N, Signature> where
...
+		true
+	}
+
+	/// Set the length of precommits received at the moment of precommiting.
+	pub fn set_precommit_idx(&mut self) -> bool {
+		self.precommit.voted_at = Some(self.precommit.votes().len());
+		println!("set precommit idx to {:?}", self.precommit.voted_at);
```
rphmeier: ditto

src/voter/voting_round.rs
```diff
@@ -273,7 +289,13 @@ impl<H, N, E: Environment<H, N>> VotingRound<H, N, E> where
 			}
 		});
 
-		prevotes.chain(precommits).collect()
+		let prevotes_len = prevotes.len();
+
+		HistoricalVotes {
+			seen: prevotes.chain(precommits).collect(),
+			prevote_idx: self.votes.prevote_idx(),
+			precommit_idx: self.votes.precommit_idx().map(|idx| idx + prevotes_len),
```
rphmeier: I don't think this logic is right. There is no guarantee that we see all prevotes before seeing any precommits. Since the estimate logic depends on both prevotes and precommits, we need an _interleaved_ list where the indices are those we've voted at.

rphmeier:
The main difficulty with this feature isn't necessarily answering the queries, but persisting them to the disk so that we can answer queries after restarts and even for old rounds.

src/voter/voting_round.rs
marcio-diaz: Ok, I thought it doesn't invalidate the votes.

marcio-diaz:
> The main difficulty with this feature isn't necessarily answering the queries, but persisting them to the disk so that we can answer queries after restarts and even for old rounds.

The persistence should be done in substrate, right? I'll make a PR there, receiving the `HistoricalVotes` and saving it in a appropriate format.

Then I'm not sure if the answering methods should be done here or in substrate. My guess is to add them in `Environment`.

rphmeier:
Ideally we can have the functionality in this crate which is generic over some trait that provides the persistence, with the implementation living in Substrate.

src/round.rs
```diff
@@ -332,6 +335,8 @@ impl<Id, H, N, Signature> Round<Id, H, N, Signature> where
 						chain,
 					)?;
 
+					self.historical_votes.seen.push(Message::Prevote(vote.clone()));
+
 					None
 				}
 				VoteMultiplicity::Equivocated(ref first, ref second) => {
```
andresilva: Don't we need to store equivocated votes as well?

src/round.rs
```diff
@@ -409,6 +414,8 @@ impl<Id, H, N, Signature> Round<Id, H, N, Signature> where
 						chain,
 					)?;
 
+					self.historical_votes.seen.push(Message::Precommit(vote.clone()));
+
 					None
 				}
 				VoteMultiplicity::Equivocated(ref first, ref second) => {
```
andresilva: Same as above.

src/voter/voting_round.rs
```diff
@@ -370,6 +377,7 @@ impl<H, N, E: Environment<H, N>> VotingRound<H, N, E> where
 					if let Some(prevote) = self.construct_prevote(last_round_state)? {
 						debug!(target: "afg", "Casting prevote for round {}", self.votes.number());
 						self.env.prevoted(self.round_number(), prevote.clone())?;
+						self.votes.set_prevoted_index();
```
andresilva: Isn't this going to race? Is it guaranteed that the next vote we process will be our own?

src/voter/voting_round.rs
```diff
@@ -255,7 +256,7 @@ impl<H, N, E: Environment<H, N>> VotingRound<H, N, E> where
 		self.best_finalized.as_ref()
 	}
 
-	/// Return all imported votes for the round (prevotes and precommits).
+	/// Return all imported votes for the round (prevotes and precommits) unordered.
```
andresilva: This is probably unused now.

src/voter/voting_round.rs
andresilva: Nevermind, I understand the semantics of this are to mark what we had seen so far when we prevoted/precommited (regardless of whether we'll eval our own vote before/after other new votes).

src/round.rs
rphmeier: At the very least, the first equivocation

marcio-diaz: Fixed.

src/round.rs
```diff
@@ -414,7 +422,9 @@ impl<Id, H, N, Signature> Round<Id, H, N, Signature> where
 						chain,
 					)?;
 
-					self.historical_votes.seen.push(Message::Precommit(vote.clone()));
+					let message = Message::Precommit(vote.clone());
```
andresilva: Is this clone necessary?

src/round.rs
```diff
@@ -414,7 +422,9 @@ impl<Id, H, N, Signature> Round<Id, H, N, Signature> where
...
+					let signed_message = SignedMessage { id: signer.clone(), signature, message };
```
andresilva: Unnecessary clone?

src/round.rs
```diff
@@ -332,13 +335,23 @@ impl<Id, H, N, Signature> Round<Id, H, N, Signature> where
...
 				}
 				VoteMultiplicity::Equivocated(ref first, ref second) => {
 					// mark the equivocator as such. no need to "undo" the first vote.
 					self.bitfield_context.equivocated_prevote(info)
 						.expect("info is instantiated from same voter set as bitfield; qed");
 
+					// Push the vote into HistoricalVotes.
+					let message = Message::Prevote(vote);
+					let signed_message = SignedMessage { id: signer.clone(), signature, message };
+					self.historical_votes.push_vote(signed_message);
```
andresilva: Maybe only add if `vote == second`? Or maybe if `vote == first || vote == second` just to be safe wrt ordering.

src/round.rs
```diff
@@ -409,13 +422,22 @@ impl<Id, H, N, Signature> Round<Id, H, N, Signature> where
...
 				}
 				VoteMultiplicity::Equivocated(ref first, ref second) => {
 					// mark the equivocator as such. no need to "undo" the first vote.
 					self.bitfield_context.equivocated_precommit(info)
 						.expect("info is instantiated from same voter set as bitfield; qed");
 
+					// Push the vote into HistoricalVotes.
+					let message = Message::Precommit(vote);
+					let signed_message = SignedMessage { id: signer.clone(), signature, message };
+					self.historical_votes.push_vote(signed_message);
```
andresilva: Same as above.

src/round.rs
andresilva: This is not needed since we early-exit after we've already seen an equivocation.

PR #67, opened 2019-05-16
Add accountable safety responses methods

marcio-diaz:
This adds the methods described in https://github.com/paritytech/finality-grandpa/issues/10#issuecomment-473811508


src/voter/voting_round.rs
```diff
@@ -256,6 +256,27 @@ impl<H, N, E: Environment<H, N>> VotingRound<H, N, E> where
 		self.best_finalized.as_ref()
 	}
 
+	/// Return all imported votes for the round (prevotes and precommits).
+	pub(super) fn votes(&self) -> Vec<SignedMessage<H, N, E::Signature, E::Id>> {
```
rphmeier: perhaps return an `impl Iterator` rather than forcing collection into a `Vec`.

src/testing.rs
```diff
@@ -216,7 +219,8 @@ impl crate::voter::Environment<&'static str, u32> for Environment {
 		_round: u64,
 		_state: RoundState<&'static str, u32>,
 		_base: (&'static str, u32),
-		_votes: &HistoricalVotes<&'static str, u32, Self::Signature, Self::Id>,
```
andresilva: Isn't the `_votes` parameter redundant now? Everything we've observed should be in `HistoricalVotes`.

src/voter/mod.rs
```diff
@@ -89,10 +89,20 @@ pub trait Environment<H: Eq, N: BlockNumberOps>: Chain<H, N> {
...
+		votes: &HistoricalVotes<H, N, Self::Signature, Self::Id>,
+	) -> Result<(), Self::Error>;
 
 	/// Note that we have precommitted in the given round.
-	fn precommitted(&self, round: u64, precommit: Precommit<H, N>) -> Result<(), Self::Error>;
+	fn precommitted(
+		&self,
+		round: u64,
+		precommit: Precommit<H, N>,
+		votes: &HistoricalVotes<H, N, Self::Signature, Self::Id>,
```
andresilva: This is passed along when prevoting/precommitting because we might need to answer a challenge in the current round?

src/lib.rs
```diff
@@ -246,6 +246,14 @@ impl<H, N: Copy> Message<H, N> {
 			Message::PrimaryPropose(ref v) => (&v.target_hash, v.target_number),
 		}
 	}
+
+	/// Return true if the message is a prevote.
+	pub fn is_prevote(&self) -> bool {
```
andresilva: Unused?

src/lib.rs
```diff
@@ -543,6 +551,14 @@ impl<H, N, S, Id> HistoricalVotes<H, N, S, Id> {
 	}
 }
 
+pub trait AccountableSafety {
```
andresilva: This is unused and will probably need to be changed once it's actually used (no particular reason it's just the way things usually go). Maybe return `impl Iterator` instead? I'm fine with adding it anyway.

src/voter/mod.rs
marcio-diaz: I thought it could end up in an inconsistent state if historical votes are not saved at prevote/precommit. I'm not sure now, but I guess it doesn't hurt.

src/lib.rs
marcio-diaz: It seems we can't return `impl Iterator` (unless I box it?) because is a trait method.  I changed it to return a `Self::Messages` for now...

src/lib.rs
marcio-diaz: I'm using it at Substrate.

src/lib.rs
```diff
@@ -495,34 +503,34 @@ pub fn process_commit_validation_result<H, N>(
 #[cfg_attr(feature = "derive-codec", derive(Encode, Decode))]
 pub struct HistoricalVotes<H, N, S, Id> {
 	seen: Vec<SignedMessage<H, N, S, Id>>,
-	prevote_idx: Option<u64>,
-	precommit_idx: Option<u64>,
+	prevote_index: Option<u32>,
```
andresilva: Not that I disagree but why change this from `u64` to `u32`?

src/lib.rs
```diff
@@ -534,27 +542,35 @@ impl<H, N, S, Id> HistoricalVotes<H, N, S, Id> {
...
 
 	/// Set the number of messages seen before precommiting.
-	pub fn set_precommited_idx(&mut self) {
-		self.precommit_idx = Some(self.seen.len() as u64)
+	pub fn set_precommited_index(&mut self) {
+		self.precommit_index = Some(self.seen.len() as u32)
 	}
 }
 
+pub trait AccountableSafety {
```
andresilva: Docs on public items.

andresilva: I need to see the substrate PR to understand what this trait is used for, seems it provides the same information we have on `HistoricalVotes`.

andresilva:
Pushed a commit to merge master and fix conflicts.

src/voter/voting_round.rs
```diff
@@ -311,6 +311,27 @@ impl<H, N, E: Environment<H, N>> VotingRound<H, N, E> where
 		self.best_finalized.as_ref()
 	}
 
+	/// Return all imported votes for the round (prevotes and precommits).
```
andresilva: This is unused in this crate and iirc `VotingRound` is not exported so this isn't used by anyone.

marcio-diaz:
Closing for the moment.

PR #70, opened 2019-07-01
Add test helpers to construct opaque types

andresilva:
The `CommitProcessingOutcome` and `CatchUpProcessingOutcome` types are opaque and it's not possible to construct them from outside this crate which is useful for testing. Used in https://github.com/paritytech/substrate/pull/2801.

### DIFF
--- a/src/voter/mod.rs
+++ b/src/voter/mod.rs
@@ -134,6 +134,19 @@ pub enum CommitProcessingOutcome {
 	Bad(BadCommit),
 }
 
+#[cfg(any(test, feature = "test-helpers"))]
+impl CommitProcessingOutcome {
+	/// Returns a `Good` instance of commit processing outcome's opaque type. Useful for testing.
+	pub fn good() -> CommitProcessingOutcome {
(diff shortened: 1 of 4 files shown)


src/voter/mod.rs
```diff
@@ -134,6 +134,19 @@ pub enum CommitProcessingOutcome {
 	Bad(BadCommit),
 }
 
+#[cfg(any(test, feature = "test-helpers"))]
+impl CommitProcessingOutcome {
+	/// Returns a `Good` instance of commit processing outcome's opaque type. Useful for testing.
+	pub fn good() -> CommitProcessingOutcome {
```
rphmeier: can we guarantee that we won't need to break API for these?

andresilva: I think we should always be able to provide some default value for this. As we add more fields to it we might need to add some different helper methods to create specific instances though. Is this what you were thinking?

rphmeier: Yeah -- mostly that I wasn't convinced that we can guarantee a default

PR #71, opened 2019-07-02
Wait for previous round estimate to be finalized

andresilva:
Before starting round R we guarantee that the estimate of round R-2 has been finalized.

### DIFF
--- a/src/voter/voting_round.rs
+++ b/src/voter/voting_round.rs
@@ -215,13 +215,13 @@ impl<H, N, E: Environment<H, N>> VotingRound<H, N, E> where
 				// either it was already finalized in the previous round
 				let finalized_in_last_round = self.env.is_equal_or_descendent_of(
 					last_round_estimate.clone(),
-					last_round_finalized.clone(),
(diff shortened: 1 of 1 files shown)


src/voter/voting_round.rs
```diff
@@ -199,11 +200,50 @@ impl<H, N, E: Environment<H, N>> VotingRound<H, N, E> where
...
+
+		// make sure that the previous round estimate has been finalized
+		let last_round_estimate_finalized = match last_round_state {
+			Some(RoundState {
+				estimate: Some((last_round_estimate, _)),
+				finalized: Some((last_round_finalized, _)),
+				..
+			}) => {
+				// either it was already finalized in the previous round
+				let finalized_in_last_round = self.env.is_equal_or_descendent_of(
```
andresilva: Can we get away with just using block numbers?

rphmeier: Yeah, that should be fine. So we are checking `last_round.estimate_number <= finalized_number` before starting `next_round`.

PR #72, opened 2019-07-04
Apply rustfmt

andresilva:
Depends on #71.

Decided to run rustfmt because I didn't have anything better to do... Even though I'd like to use a formatting tool I don't have a strong opinion on whether we should stick with this. Overall I think the formatting looks OK.

I'm using some unstable rustfmt settings so it requires the nightly version (`cargo +nightly fmt`).


andresilva:
I don't mean to start a flamewar, I was curious how it would look.

PR #76, opened 2019-07-10
Fix no std build

andresilva:
`process_commit_validation_result` should only be compiled under `std` environment since it uses the voter module (which isn't compiled under `no-std`).

Also add a CI build for `no-std`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,7 +26,6 @@
 //! The work for actually casting votes is done in the `voter` module.
 
 #![cfg_attr(not(feature = "std"), no_std)]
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -26,7 +26,6 @@
 //! The work for actually casting votes is done in the `voter` module.
 
 #![cfg_attr(not(feature = "std"), no_std)]
-#![cfg_attr(not(feature = "std"), feature(alloc))]
```
andresilva: Not needed anymore in Rust 1.36.

PR #79, opened 2019-07-25
Fix grandpa no-std

marcio-diaz:
This add a couple of guards to allow importing structs (`Message`, `Equivocation`) into no-std.

It may be enough for my use case in Substrate (at least for equivocations).

But the plan is to make follow up PRs implementing no-std for each guarded module (and removing the corresponding guard).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,9 +8,9 @@ repository = "https://github.com/paritytech/finality-grandpa"
 edition = "2018"
 
 [dependencies]
-parking_lot = "0.9"
 futures = "0.1"
 log = "0.4"
+parking_lot = { version = "0.9", optional = true }
 parity-codec = { version = "4.1", optional = true, default-features = false, features = ["derive"] }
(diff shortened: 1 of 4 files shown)


marcio-diaz:
> I don't understand this, isn't this code compilable under no-std? I thought everything supported no-std except for the voter module. 

It's using the std library somewhere inside the modules and we are not guarding it.

> Why are we restricting stuff to only be std?

Because anyways (as it is) only works when importing into an environment that supports std library.
The guards are just make it explicit.

The modules are filtered in the branch `marcio/grandpa-primitives` already, I just need to clean it a bit.

marcio-diaz:
@andresilva I removed the guards in the main file and added some to the particular modules, so now you can take a look again. 

The main problem was that `parking_lot` doesn't seems to support `no_std` (obviously?). I handled it in the trivial way, but maybe other option is to mock `RawRwLock`.

PR #81, opened 2019-08-16
Update to stables futures

tomaka:
Updates the crate to use futures from the standard library rather than the `futures` crate.

### DIFF
--- a/src/voter/mod.rs
+++ b/src/voter/mod.rs
@@ -25,14 +25,16 @@
 //!  votes will not be pushed to the sink. The protocol state machine still
 //!  transitions state as if the votes had been pushed out.
 
-use futures::prelude::*;
-use futures::sync::mpsc::{self, UnboundedReceiver};
+use futures::{prelude::*, ready};
+use futures::channel::mpsc::{self, UnboundedReceiver};
 #[cfg(feature = "std")]
 use log::trace;
 
 use std::collections::VecDeque;
 use std::hash::Hash;
+use std::pin::Pin;
 use std::sync::Arc;
+use std::task::{Context, Poll};
 
 use crate::round::State as RoundState;
 use crate::{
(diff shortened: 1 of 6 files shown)


src/voter/mod.rs
```diff
@@ -302,13 +304,13 @@ pub struct RoundData<Id, Timer, Input, Output> {
 	pub outgoing: Output,
 }
 
-struct Buffered<S: Sink> {
+struct Buffered<S, I> {
```
tomaka: Note: the signature of the `Sink` trait has changed. The item is now a parameter of the trait instead of an associated type.

src/voter/mod.rs
```diff
@@ -888,17 +884,17 @@ mod tests {
...
-					.take_while(|&(_, n, _)| Ok(n < 6))
-					.for_each(|_| Ok(()))
-			});
+					.take_while(|&(_, n, _)| future::ready(n < 6))
+					.for_each(|_| future::ready(()))
+			}).collect::<Vec<_>>();
 
-			::futures::future::join_all(finalized_streams).map(|_| signal.fire())
-		})).unwrap();
+			threads_pool.spawn_ok(routing_task.map(|_| ()));
```
tomaka: Note: Because the behaviour of `join_all` changed, I had to adjust the order of initialization, otherwise the test wouldn't work (blocked at round 1 with futures not being notified). This is a bit concerning, but I couldn't find the origin of the issue.

src/voter/mod.rs
rphmeier: Nice - it didn't exist when I wrote this code.

tomaka:
Ping!

andresilva:
@tomaka sorry for being a shitty maintainer :crying_cat_face:

rphmeier:
Yup I also gave this another look after the ping, also worried by the `join_all` thing.

src/voter/mod.rs
romanb: From what I can tell this is not related to `join_all`. If the `routing_task` future is spawned before the first `Voter` is created (i.e. the case where the test stalls / hangs), the `NetworkRouting` future is polled "immediately" before the first voter is created. It then returns `Poll::Pending`, which it always does, since it relies on task wakeup from the `UnboundedReceiver`s (i.e. when messages are sent). However, there are no voters and hence no `VotingRound` yet, so the current task is only scheduled to  be polled again if a message is sent to the `GlobalMessageNetwork`, which doesn't happen in this test. So the `NetworkRouting` is never polled again, even after the first round with voters is created, resulting in the messages never being dispatched.

In all other tests except one (`skips_to_latest_round_after_catch_up`), the `routing_task` is only spawned after the first voter is created. In `skips_to_latest_round_after_catch_up`, the it is spawned earlier, but this test then calls `network.send_message` after a voter is set up, i.e. sends a message on the `GlobalMessageNetwork`, resulting in the `NetworkRouting` to be polled again.

The reason this worked previously is most likely due to the semantics of `block_on_all` in tokio, which results in the spawned "inner" `NetworkRouting` future to be polled again after the "outer" future resolved, at which point the voters have been set up and a voting round has been created.

So it seems to me that spawning the `routing_task` only after at least one voter has been created is not only necessary now but arguably more correct, at least the way the `NetworkRouting` is currently set up.

romanb:
See https://github.com/paritytech/finality-grandpa/pull/81/files#r330677305 regarding the mentioned concern.

tomaka:
Should we merge this?
I know that GrandPa's code is very sensible and we should be extra careful, but on the other hand I think this PR is quite straight-forward and I don't really see how a bug could be introduced just by transitioning to stable futures.

src/voter/mod.rs
andresilva: Thanks for looking into this and for the detailed explanation @romanb.

Issue #83, opened 2019-08-19
Replace hashmap_core with hashbrown


andresilva:
https://crates.io/crates/hashbrown

andresilva:
Closed by #84.

PR #84, opened 2019-08-20
Replace hashmap_core with hashbrown

koushiro:
Signed-off-by: koushiro <koushiro.cqx@gmail.com>

related issue: #83

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ log = "0.4"
 parking_lot = { version = "0.9", optional = true }
 parity-scale-codec = { version = "1.0.3", optional = true, default-features = false, features = ["derive"] }
 num = { package = "num-traits", version = "0.2", default-features = false }
-hashmap_core = { version = "0.1.10", default-features = false }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -13,7 +13,7 @@ log = "0.4"
 parking_lot = { version = "0.9", optional = true }
 parity-scale-codec = { version = "1.0.3", optional = true, default-features = false, features = ["derive"] }
 num = { package = "num-traits", version = "0.2", default-features = false }
-hashmap_core = { version = "0.1.10", default-features = false }
+hashbrown = { version = "0.6" }
```
andresilva: We need to compile with `default-features = false`. `finality-grandpa` is `no-std` compatible.

koushiro: `hashbrown` supports no_std by default.

andresilva: Oh, I didn't know that. Thanks :)

andresilva:
Thanks for the PR @koushiro! :)

PR #86, opened 2019-09-30
Don't use more core modules than Substrate does without "std".

### LINKED ISSUE #85: Remove std::fmt dependency without "std" feature
This crate currently uses `std::fmt` even when the "std" feature is disabled. See https://github.com/paritytech/finality-grandpa/blob/master/src/lib.rs#L56

I'm working on a Substrate module that requires this crate and in SRML, it seems the recommendation is to not use `std::fmt` when "std" is off. Instead, SRML modules can make use of the `MaybeDebug`, `MaybeDisplay`, etc [traits defined in the sr_primitives](https://github.com/paritytech/finality-grandpa/blob/master/src/lib.rs#L56) crate.

Would it be acceptable to add `sr_primitives` as a dependency of this? Or should we copy the small amount of code defining the traits into this project? I don't believe this would lead to any sort of incompatibility issues.

jimpo:
This prepares the crate to be used by a Substrate module without requiring more core crate modules than SRML uses already. The main difficulty is in removing the use of core::fmt.

Fixes #85.

### DIFF
--- a/src/bitfield.rs
+++ b/src/bitfield.rs
@@ -28,19 +28,18 @@
 //! Bitfields on regular vote-nodes will tend to be live, but the equivocating
 //! bitfield will be mostly empty.
 
-use std::fmt;
-
 #[cfg(feature = "std")]
 use parking_lot::RwLock;
 
 #[cfg(feature = "std")]
 use std::sync::Arc;
 
-use crate::collections::Vec;
+use crate::std::{self, vec::Vec};
 use crate::voter_set::VoterInfo;
 
 /// Errors that can occur when using the equivocation weighting tools.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, PartialEq, Eq)]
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -53,28 +77,15 @@ mod bridge_state;
 #[cfg(test)]
 mod testing;
 
+#[cfg(feature = "std")]
 use std::fmt;
```
andresilva: we can probably remove this import and just refer to `std::fmt` where this is used below.

src/bitfield.rs
```diff
@@ -28,6 +28,7 @@
 //! Bitfields on regular vote-nodes will tend to be live, but the equivocating
 //! bitfield will be mostly empty.
 
+#[cfg(feature = "std")]
 use std::fmt;
```
andresilva: Let's just remove it and use `std::fmt` below.

PR #87, opened 2019-10-21
bump version to v0.9.1

andresilva:
Need to release latest master since it includes some fixes for building under no-std environment (needed for https://github.com/paritytech/substrate/pull/3868).

The only changes included from v0.9.0 are:
- https://github.com/paritytech/finality-grandpa/pull/84
- https://github.com/paritytech/finality-grandpa/pull/86

Neither of them seems breaking.

### DIFF
--- a/src/round.rs
+++ b/src/round.rs
@@ -299,6 +299,7 @@ impl<Id, H, N, Signature> Round<Id, H, N, Signature> where
 		self.round_number
 	}
 
+	#[allow(unused)]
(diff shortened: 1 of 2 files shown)


src/round.rs
```diff
@@ -299,6 +299,7 @@ impl<Id, H, N, Signature> Round<Id, H, N, Signature> where
 		self.round_number
 	}
 
+	#[allow(unused)]
```
andresilva: When compiling under `no_std` this is unused since the voter is not included.

PR #88, opened 2019-11-09
launch voter with last round in background

rphmeier:


src/voter/voting_round.rs
```diff
@@ -230,8 +232,11 @@ impl<H, N, E: Environment<H, N>> VotingRound<H, N, E> where
 			_ => false,
 		};
 
+		self.log_participation();
+
 		// the previous round estimate must be finalized
 		if !last_round_estimate_finalized {
+			trace!("round {} completable but estimate not finalized.", self.round_number());
```
andresilva:
```suggestion
			trace!(target: "afg", "round {} completable but estimate not finalized.", self.round_number());
```

andresilva:
Let's keep this open until we actually figure out what happened.

src/voter/mod.rs
```diff
@@ -82,6 +84,11 @@ pub trait Environment<H: Eq, N: BlockNumberOps>: Chain<H, N> {
 		Self::Out,
 	>;
 
+	/// Return the final round state of the given round number's parent, if stored.
```
andresilva:
```suggestion
	/// Return the latest round state of the given completed round, if stored.
```

src/voter/mod.rs
```diff
@@ -82,6 +84,11 @@ pub trait Environment<H: Eq, N: BlockNumberOps>: Chain<H, N> {
...
+	fn round_parent_state(&self, _round: u64) -> Option<RoundState<H, N>> {
```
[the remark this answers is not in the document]
andresilva: Why not make it explicit that the given round must have been previously completed and also drop the parent thing from the interface (we can always call this on `round - 1`).

src/voter/mod.rs
```diff
@@ -361,6 +368,51 @@ type FinalizedNotification<H, N, E> = (
...
+	H: Hash + Clone + Eq + Ord + ::std::fmt::Debug,
+	N: Copy + BlockNumberOps + ::std::fmt::Debug,
+{
+	let last_round_tracker = crate::round::Round::new(crate::round::RoundParams {
+		voters: voters,
+		base: last_round_base,
+		round_number: last_round_number,
+	});
+
+	let last_round_parent_state = env.round_parent_state(last_round_number)
```
andresilva: Do we actually need this? The previous round state is only necessary for voting and to make sure we don't start `R + 1` while the estimate for `R - 1` hasn't been finalized. In this case we are spawning `R - 1` and `R`, we won't vote on `R - 1` (we must have done so already previously otherwise it wouldn't be completed) and we only need to make sure that `R` has access to `R - 1`'s state to track any estimate change.

src/voter/mod.rs
```diff
@@ -361,6 +368,51 @@ type FinalizedNotification<H, N, E> = (
 	Commit<H, N, <E as Environment<H, N>>::Signature, <E as Environment<H, N>>::Id>,
 );
 
+// instantiates the last round, to be backgrounded until its estimate is finalized.
```
andresilva:
```suggestion
// Instantiates the given last round, to be backgrounded until its estimate is finalized.
```

src/voter/mod.rs
```diff
@@ -361,6 +368,51 @@ type FinalizedNotification<H, N, E> = (
...
+//
+// this round must be comletable based on the passed votes (and if not, `None` will be returned)
```
andresilva:
```suggestion
// This round must be completable based on the passed votes (and if not, `None` will be returned),
```

src/voter/mod.rs
```diff
@@ -361,6 +368,51 @@ type FinalizedNotification<H, N, E> = (
...
+// but it may be the case that there are some more votes to propagate in order to push
+// the estimate backwards.
```
andresilva:
```suggestion
// the estimate backwards and conclude the round (i.e. finalize its estimate).
```

src/voter/mod.rs
rphmeier: I think i tried it without; the problem is that the round concludes immediately when the grandparent state is `None`.

src/voter/mod.rs
```diff
@@ -1168,4 +1241,105 @@ mod tests {
...
+					SignedMessage {
+						message: crate::Message::Precommit(crate::Precommit {
+							target_hash: "D",
+							target_number: 5,
+						}),
+						signature: Signature(5),
+						id: local_id,
+					},
+				],
+				last_finalized,
```
andresilva: This will be `(Genesis, 0)` if I'm not mistaken, but the grandparent round finalized `(C, 4)`. We should make a test where we actually need to have the background running in order to progress to the latest round (i.e. we need votes on the last round to push the estimate backwards).

rphmeier:
(only the coverage job is running rn)

PR #91, opened 2019-11-18
do not compute estimate until round could possibly be completable

rphmeier:
(updated after Roman's comment below)

We do not compute the estimate until the round could possibly be completable, which is when no new leaf nodes can get a threshold amount of precommits. Until that point, the estimate of what has been finalized is probably not useful, although technically equal to the `prevote_ghost`. This is a relaxation of the pre-existing rule.

### DIFF
--- a/src/round.rs
+++ b/src/round.rs
@@ -798,7 +798,26 @@ mod tests {
 		).unwrap();
 
 		assert_eq!(round.prevote_ghost, Some(("E", 6)));
+		assert_eq!(round.estimate(), None);
+		assert!(!round.completable());
+
+		round.import_precommit(
+			&chain,
+			Precommit::new("E", 6),
+			"Alice",
+			Signature("Alice"),
+		).unwrap();
+
+		round.import_precommit(
+			&chain,
+			Precommit::new("E", 6),
+			"Bob",
+			Signature("Bob"),
+		).unwrap();
+
+		assert_eq!(round.prevote_ghost, Some(("E", 6)));
 		assert_eq!(round.estimate(), Some(&("E", 6)));
+		assert!(round.completable());
 	}
 
 	#[test]
--- a/src/round.rs
+++ b/src/round.rs
@@ -640,6 +640,9 @@ impl<Id, H, N, Signature> Round<Id, H, N, Signature> where
 				possible_to_precommit,
 			);
 		} else {
+			// although technically the estimate would be equal to the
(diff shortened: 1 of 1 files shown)


romanb:
> [..] having an estimate is not a useful notion until we have threshold precommit-weight.

Is that true though? Take `n = 5` with `f = 1` and `threshold = 4`, now if 3 votes have been seen for some block `A` (not a descendant of `A`), then it is impossible for any descendant of `A` to obtain supermajority 4, hence the estimate is monotonically decreasing from `A` and the round completable.

More generally, there is the following comment a bit above the code in the changeset relating to the condition on when to (re-)compute the estimate:

>// until we have threshold precommits, any new block could get supermajority
> // precommits because there are at least f + 1 precommits remaining and then
> // f equivocations.

However, this only seems to hold when `n = 3f + 1` in the natural numbers, i.e. for `n = 4, 7, 10, ...`, but not for any other `n` - that is, as long as we don't have threshold precommits there is always _at least_ `2f + 1` vote weight for any new block to achieve, but that is only enough for supermajority when `n = 3f + 1`, not if `n > 3f + 1`. As far as I can tell, truly _any_ new block  can still get supermajority only as long as `[num-votes-remaining] + f >= threshold`, and thus a condition for when to compute the estimate because it _may be_ already monotonically decreasing (and thus the round completable) would be `[num-votes-remaining] + f < threshold`. I hope that makes sense.

rphmeier:
Thanks Roman - that all seems correct with one correction

> As far as I can tell, truly any new block can still get supermajority only as long as [num-votes-remaining] + f >= threshold, and thus a condition for when to compute the estimate because it may be already monotonically decreasing

It should rather be `[num-votes-remaining] + f - e >= threshold`, where `e` is the amount of weight which has already equivocated.

I'll adjust the PR accordingly.

rphmeier:
Thanks Roman. I'm updating the PR. Instead, I don't update the estimate because it is "not useful until the round could possibly be completable". It is worth noting that the paper defines the impossibility condition of a child gaining supermajority based on having at least `2f + 1` votes.

andresilva:
Thanks for the explanation and example @romanb.

rphmeier:
I chatted to Alistair about this: the impossibility threshold can be chosen in `2f + 1 ... n - f`. Any is valid (but IMO the 2f + 1 is most correct)

PR #92, opened 2019-11-18
Add a concluded function to environment

rphmeier:
...called when a round's estimate is finalized and the round will no longer be polled.

On top of #88

### DIFF
--- a/src/testing.rs
+++ b/src/testing.rs
@@ -243,6 +243,16 @@ pub mod environment {
 			Ok(())
 		}
 
+		fn concluded(
+			&self,
+			_round: u64,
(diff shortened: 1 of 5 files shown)


src/testing.rs
```diff
@@ -135,7 +135,7 @@ pub mod environment {
 	use futures::prelude::*;
 	use futures::sync::mpsc::{self, UnboundedReceiver, UnboundedSender};
 	use parking_lot::Mutex;
-	use std::collections::HashMap;
+	use std::collections::{HashMap, HashSet};
```
andresilva:
```suggestion
	use std::collections::HashMap;
```

PR #93, opened 2019-11-21
VoterSet and Bitfield Refactoring

romanb:
This is a proposal for a refactoring (or rather rewrite) of the `voter_set` and `bitfield` modules alongside smaller related changes in the `round` and `vote_graph`, targeting the `v0.11.0` branch (i.e. based on https://github.com/paritytech/finality-grandpa/pull/81).

Below is a rough outline of the main changes.

## Bitfield

  * Simplify definition, as `Bitfield::Blank` and `LiveBitfield(vec![])`
    are equivalent.
  * Dynamically-sized with lazy allocation makes for a more ergonomic
    API and keeps bitfields smaller with larger voter sets. The reasoning
    here is the following: If there are few voters, most votes
    are likely on the same or very few nodes / blocks but in that case bitfields
    are almost never resized anyway. As there are more and more voters,
    votes are likely spread across more and more nodes / blocks and thus not
    all nodes need a full-sized bitfield (whose size grows with the voter set).
  * Avoid allocations when merging bitfields. Previously merging would
    always allocate a new bitfield, now the other bitfield is merged
    into `self` and only allocates if `self` needs to be resized.
  * Avoid allocations when recording votes in the vote graph and computing weights.
    Previously both recording of new votes and computing weights
    involved creating and merging bitfields. Now incorporating a
    vote into an existing vote-node is a `set_bit` operation while
    computing the vote weight on a node is based on a bitfield-merging
    iterator that does not allocate.
  * Better tests.

## VoterSet

  * Make it impossible to construct inconsistent sets as well as
    empty sets. Both were previously possible.
  * Better tests.

## Related changes

  * Introduce a `weights` module for encapsulating vote-weight
    arithmetic behind newtypes.
  * Introduce a `round::context` module for encapsulating the
    context of a round in which vote weights are computed,
    consisting of the voter set and recorded equivocations.
    This combines some functionality from the previous `bitfield`
    module and some from the parent `round` module.

### DIFF
--- a/src/round/context.rs
+++ b/src/round/context.rs
@@ -17,7 +17,6 @@
 
 use crate::bitfield::{Bitfield, Bit1};
 use crate::voter_set::{VoterSet, VoterInfo};
-use crate::std::hash::Hash;
 use crate::std::ops::AddAssign;
 use crate::weights::VoteWeight;
 
@@ -26,12 +25,12 @@ use super::Phase;
 /// The context of a `Round` in which vote weights are calculated.
 #[cfg_attr(feature = "std", derive(Debug))]
 #[cfg_attr(test, derive(Clone))]
-pub struct Context<T: Eq + Hash> {
+pub struct Context<T: Ord + Eq> {
 	voters: VoterSet<T>,
 	equivocations: Bitfield,
 }
 
-impl<T: Eq + Hash> Context<T> {
+impl<T: Ord + Eq> Context<T> {
 	/// Create a new context for a round with the given set of voters.
 	pub fn new(voters: VoterSet<T>) -> Self {
 		Context {
@@ -101,7 +100,7 @@ impl Vote {
 	/// if it is contained in that set.
 	fn voter<'a, Id>(&'a self, vs: &'a VoterSet<Id>) -> Option<(&'a Id, &'a VoterInfo)>
 	where
-		Id: Eq + Hash
+		Id: Eq + Ord
 	{
 		vs.nth(self.bit.position / 2)
 	}
@@ -141,7 +140,7 @@ impl AddAssign<&Vote> for VoteNode {
 /// of vote-bits in the context of the given set of voters.
 fn weight<V, I>(bits: I, voters: &VoterSet<V>) -> VoteWeight
 where
-	V: Eq + Hash,
+	V: Eq + Ord,
 	I: Iterator<Item = Bit1>
 {
 	let mut total = VoteWeight(0);
--- a/src/round/context.rs
+++ b/src/round/context.rs
@@ -1,3 +1,16 @@
+// Copyright 2019 Parity Technologies (UK) Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
 
 //! The context of a GRANDPA round tracks the set of voters
 //! and equivocations for the purpose of computing vote weights.
--- a/src/round/context.rs
+++ b/src/round/context.rs
@@ -23,7 +23,7 @@ use crate::weights::VoteWeight;
 use super::Phase;
 
 /// The context of a `Round` in which vote weights are calculated.
-#[cfg_attr(feature = "std", derive(Debug))]
+#[cfg_attr(any(feature = "std", test), derive(Debug))]
 #[cfg_attr(test, derive(Clone))]
 pub struct Context<T: Ord + Eq> {
 	voters: VoterSet<T>,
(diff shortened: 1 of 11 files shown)


rphmeier:
Looks good at a skim - going to give it a more in-depth review soon. Thanks!

src/round/context.rs
```diff
@@ -0,0 +1,224 @@
+
+//! The context of a GRANDPA round tracks the set of voters
```
rphmeier: missing license headers

src/vote_graph.rs
```diff
@@ -87,17 +87,17 @@ pub struct VoteGraph<H: Hash + Eq, N, V> {
 
 impl<H, N, V> VoteGraph<H, N, V> where
 	H: Hash + Eq + Clone + Ord + Debug,
-	V: AddAssign + Default + Clone + Debug,
+	V: for<'a> AddAssign<&'a V> + Default + Clone + Debug,
 	N: Copy + Debug + BlockNumberOps,
 {
 	/// Create a new `VoteGraph` with base node as given.
-	pub fn new(base_hash: H, base_number: N) -> Self {
+	pub fn new(base_hash: H, base_number: N, base_node: V) -> Self {
```
rphmeier: do we have any use-case where we are initializing this to non-default? The `Default` ~ `Zero` property is important for a lot of the GHOST-stuff and this may discourage that.

src/voter_set.rs
```diff
@@ -12,148 +12,241 @@
...
+
+		// Populate the voter set, thereby calculating the total weight.
+		let mut voters = HashMap::with_capacity(weights.size_hint().0);
+		let mut total_weight = 0u64;
+		for (id, weight) in weights {
+			if let Some(w) = NonZeroU64::new(weight) {
+				// Prevent construction of inconsistent voter sets by checking
+				// for weight overflow (not just in debug mode). The protocol
+				// should never run with such voter sets.
+				total_weight = total_weight.checked_add(weight).expect("Voter weight overflow");
```
rphmeier: the general policy we have for panickers is to avoid unless it's provable they can't trigger. Could we return `None` or an error variant here?

src/voter_set.rs
```diff
@@ -12,148 +12,241 @@
...
-	}
+		let total_weight = VoterWeight::new(total_weight).expect("voters nonempty; qed");
 
-	// Get voter by index.
-	pub fn voter_by_index(&self, idx: usize) -> &(Id, u64) {
-		&self.voters[idx]
-	}
+		// Establish the total order on the set based on the voter IDs.
+		let mut order = voters.keys().cloned().collect::<Vec<_>>();
+		order.sort_unstable();
```
rphmeier: All nodes need to agree on the canonical ordering (for unanimous primary selection), so a stable sort is safest. I don't think different versions of `core` are required to use the same unstable sorting algorithm, which could lead to consensus confusion.

src/voter_set.rs
```diff
@@ -12,148 +12,241 @@
...
+	pub fn contains(&self, id: &Id) -> bool {
+		self.voters.contains_key(id)
+	}
 
-		let mut voters = Vec::with_capacity(lower);
-		let mut weights = HashMap::with_capacity(lower);
+	/// Get the nth voter in the set, modulo the size of the set,
+	/// as per the associated total order.
+	pub fn nth_mod(&self, n: usize) -> (&Id, &VoterInfo) {
+		self.nth(n % self.order.len()).unwrap()
```
rphmeier: `unwrap`s are definitely not allowed

rphmeier:
I've been fuzz-testing the old implementation for the last 5 days on the `rh-fuzzing` branch without exposing any issues. I'm worried that the rewrite opens new opportunities for logic errors, particularly in equivocation counting, GHOST, and completability.

src/voter_set.rs
romanb: Missed that one, thanks, it was not intentional.

src/voter_set.rs
romanb: But the difference between a stable and unstable sort only related to equal elements. There are no two equal elements in `order` by its definition from the line above. What am I missing?

src/voter_set.rs
romanb: Sure, if you prefer that.

src/vote_graph.rs
romanb: This was primarily motivated by the fact that making use of a type parameter constraint (i.e. trait bound) in a method that whereby the type parameter appears nowhere in the method's arguments necessarily leaves the parameter uninstantiated in expressions of the form
```rust
let graph = VoteGraph::new(...);
```
and thus makes type inference harder. Previously `V` could then be instantiated by the type system through observing subsequent occurrences of `graph.insert(...)`. Now the parameter of `insert` is some `W` such that `V: for<'a> AddAssign<&'a W>`, so `V` is still undetermined. It can of course be solved by giving an explicit type ascription to expressions such as the above (`let graph: ... = `), it just seemed to me that passing the base node in the constructor is preferable as it is more explicit and avoids this type ambiguity.

Generally the `VoteGraph` has no clue what `V::default()` produces, i.e. there was and is no `Default` ~ `Zero` property. However, it is certainly true that this allows constructing a `VoteGraph` with a base node that is different from what the default instance returns and the latter is used inside the `VoteGraph` to construct intermediate new nodes on demand. If that is undesirable, I can certainly go with the extra type annotations.

romanb:
> I would rather not merge this until it has at least the same degree of testing. [..]

I'm fine with waiting until the fuzz-testing is in master and the `v0.11` and this branch subsequently rebased on top.

romanb:
If there is still interest in this PR, I can eventually at some point rebase it over the current master. Let me know.

rphmeier:
Yep, there's still interest! Please do rebase it.

rphmeier:
Could you re-target the PR again against master as well?

romanb:
> Could you re-target the PR again against master as well?

Of course.

PR #94, opened 2019-11-21
build: fix hashbrown features under no-std

andresilva:
The default set of features for hashbrown is: "ahash", "ahash-compile-time-rng", "inline-more".
Without "ahash" the HashMap API isn't implemented, so no-std build was broken by #90.


andresilva:
We should probably just replace these maps with BTrees, especially for no-std builds where we don't want to introduce a runtime DoS vector.

PR #96, opened 2019-11-25
if we cannot prevote, we stop voting in the round

rphmeier:


src/voter/voting_round.rs
```diff
@@ -452,9 +452,16 @@ impl<H, N, E: Environment<H, N>> VotingRound<H, N, E> where
 						self.env.prevoted(self.round_number(), prevote.clone())?;
 						self.votes.set_prevoted_index();
 						self.outgoing.push(Message::Prevote(prevote));
+						self.state = Some(State::Prevoted(precommit_timer));
+					} else {
+						// when we can't construct a prevote, we shouldn't
```
andresilva: Maybe add a log message here under debug?

rphmeier: `construct_prevote` is already logging, otherwise i would have added one.

PR #99, opened 2019-11-27
derive debug under no-std for some types

andresilva:
These types are part of parameters used in the runtime for equivocation reporting and therefore need to implement `Debug`. I kept the guards for `std` for the rest of the types should we remove them now that we support `Debug` under `no-std`?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -85,9 +85,7 @@ pub mod fuzz_helpers;
 use parity_scale_codec::{Encode, Decode};
 
 /// A prevote for a block and its ancestors.
-#[derive(Clone, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


rphmeier:
AFAIK the reason we usually don't generate Debug impls in `no-std` is not because it's unsupported but to keep the code size down. Am I interpreting your first sentence as `Debug` being a required bound in Runtime? If so, happy to merge

tomusdrw:
@rphmeier Yes, substrate is now using a custom `RuntimeDebug` derive, which basically generates an empty `Debug` (exactly to avoid blowing up the wasm blob size; but note that unused Debug impls are easily removed by the compiler), but we keep the `Debug` bound for simplicity.

PR #89, opened 2019-11-09
Revert "Update to stables futures"

rphmeier:
Reverts paritytech/finality-grandpa#81

PR #100, opened 2019-11-28
Upgrade to stable futures (#2).

romanb:
This is a new version of https://github.com/paritytech/finality-grandpa/pull/81 which was later reverted in https://github.com/paritytech/finality-grandpa/pull/89. Compared to that PR, the following has changed:

  * Rebased on top of the latest changes in `master`.
  * Updated to `futures-0.3.1` and `futures_timer-2.0.2`
  * The voter tests have been changed to use a `LocalPool` instead of a `ThreadPool`. The introduction of the `ThreadPool` in the tests in https://github.com/paritytech/finality-grandpa/pull/81 caused at least one test to be subject to a race condition (`import_commit_for_any_round`). While that can of course be addressed in the test, using a `LocalPool` is much closer to the current tokio behaviour in the tests, where all futures spawned within the `lazy` block are only polled once the `lazy` block has completed, since these spawns happen in the context of a current-thread executor (although this was not even guaranteed by tokio). In any case, it makes the tests less fragile to reordering of statements, hence I went with the `LocalPool` to simplify the tests.

To not introduce too many changes (and because I want to rebase https://github.com/paritytech/finality-grandpa/pull/93 in turn on top of this PR), I did not try to make use of async/await anywhere.

### DIFF
--- a/src/voter/past_rounds.rs
+++ b/src/voter/past_rounds.rs
@@ -22,16 +22,18 @@
 //!   - Passing it any validated commits (so backgrounded rounds don't produce conflicting ones)
 
 #[cfg(feature = "std")]
-use futures::try_ready;
+use futures::ready;
 use futures::prelude::*;
 use futures::stream::{self, futures_unordered::FuturesUnordered};
 use futures::task;
-use futures::sync::mpsc;
+use futures::channel::mpsc;
 #[cfg(feature = "std")]
 use log::{trace, debug};
 
 use std::cmp;
 use std::collections::HashMap;
+use std::pin::Pin;
+use std::task::{Context, Poll};
 
 use crate::{Commit, BlockNumberOps};
 use super::Environment;
@@ -46,7 +48,7 @@ struct BackgroundRound<H, N, E: Environment<H, N>> where
 	N: Copy + BlockNumberOps + ::std::fmt::Debug,
 {
 	inner: VotingRound<H, N, E>,
-	task: Option<task::Task>,
+	waker: Option<task::Waker>,
 	finalized_number: N,
 	round_committer: Option<RoundCommitter<H, N, E>>,
 }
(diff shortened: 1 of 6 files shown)


romanb:
A note on the test failure in [this run](https://travis-ci.org/paritytech/finality-grandpa/jobs/618296626?utm_medium=notification&utm_source=github_status) of `broadcast_commit_only_if_newer`: This is actually not new. I reproduced the same sporadic error on master earlier. To witness:

```
thread 'voter::tests::broadcast_commit_only_if_newer' panicked at 'assertion failed: res.is_err()', src/voter/mod.rs:1118:6
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
test voter::tests::broadcast_commit_only_if_newer ... FAILED
```

I can either take a closer look at that in the context of this PR or create a ticket for it, whatever is preferable.

rphmeier:
A separate ticket for the test failure sounds good, since it's not caused by this PR.

src/voter/past_rounds.rs
```diff
@@ -97,35 +99,40 @@ impl<H, N, E: Environment<H, N>> Future for BackgroundRound<H, N, E> where
...
-			Ok(Async::Ready(BackgroundRoundChange::Concluded(self.round_number())))
+			Poll::Ready(Ok(BackgroundRoundChange::Concluded(self.round_number())))
 		} else {
-			Ok(Async::NotReady)
+			Poll::Pending
 		}
 	}
 }
 
+impl<H, N, E: Environment<H, N>> Unpin for BackgroundRound<H, N, E> where
```
rphmeier: I'm surprised this isn't an auto impl

rphmeier:
To avoid any divergence between GRANDPA master and whatever Substrate is pointing to, I'd prefer to have a PR to Substrate ready to go before merging this. Pierre also started on one of those in https://github.com/paritytech/substrate/pull/3909

expenses:
I might have a go at updating Pierre's PR.

PR #104, opened 2020-01-14
Changes needed for substrate#4612

expenses:
See https://github.com/paritytech/substrate/pull/4612. This merges in https://github.com/paritytech/finality-grandpa/pull/100, but also:

* Changes `rand` to be dependent on the `test-helpers` feature, which means that the crate actually compiles with this feature turned on. This does mean that you have to run `cargo test --feature test-helpers` to test it though.
* Adds an `Unpin` impl for a few types.


rphmeier:
> Changes rand to be dependent on the test-helpers feature, which means that the crate actually compiles with this feature turned on. This does mean that you have to run cargo test --feature test-helpers to test it though.

That is pretty unergonomic. Can you explain the rationale?

expenses:
@rphmeier I've added rand back as a `dev-dependency` now, so it should just work.

Cargo.toml
```diff
@@ -8,20 +8,20 @@ repository = "https://github.com/paritytech/finality-grandpa"
 edition = "2018"
 
 [dependencies]
-futures = "0.1"
+futures = "0.3.1"
+futures-timer = "2.0.2"
```
rphmeier: does it still compile in `no-std`? I would have thought we needed `no-default-features` on these futures crates.

expenses:
@rphmeier I need to test that.

Cargo.toml
expenses: Fixed.

expenses:
@rphmeier I think we're good to go here. Might be best to merge in https://github.com/paritytech/finality-grandpa/pull/100 first if you're happy with that.

rphmeier:
Solve conflicts, then merge here. Thanks!

rphmeier:
We'll need to publish a new GRANDPA version; cc @andresilva

PR #106, opened 2020-01-14
Fix sending of finality notification

andresilva:
If the voter receives enough votes to finalize before having precommitted, it is possible that it will miss sending finality notifications.

I wrote a test for this:
- Use a voter set with 4 voters
- Spin up 1 voter
- Send it 3 prevotes and 3 precommits immediately
- The node should finalize a block without having precommitted first
- No finality notification is triggered

Unfortunately the test doesn't work without some changes to the code, namely commenting this line https://github.com/paritytech/finality-grandpa/blob/master/src/voter/voting_round.rs#L183. For the issue to happen the votes required to reach finality must be processed immediately after the voter tries to prevote/precommit, this isn't easy to simulate in a test environment.

### DIFF
--- a/src/voter/voting_round.rs
+++ b/src/voter/voting_round.rs
@@ -15,13 +15,15 @@
 //! Logic for voting and handling messages within a single round.
 
 #[cfg(feature = "std")]
-use futures::try_ready;
+use futures::ready;
 use futures::prelude::*;
-use futures::sync::mpsc::UnboundedSender;
+use futures::channel::mpsc::UnboundedSender;
 #[cfg(feature = "std")]
 use log::{trace, warn, debug};
 
+use std::pin::Pin;
 use std::sync::Arc;
+use std::task::{Context, Poll};
 
 use crate::round::{Round, State as RoundState};
 use crate::{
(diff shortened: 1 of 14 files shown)


src/voter/voting_round.rs
```diff
@@ -606,17 +606,23 @@ impl<H, N, E: Environment<H, N>> VotingRound<H, N, E> where
...
+		// this is a workaround that ensures when we re-instantiate the voter after
+		// a shutdown, we never re-create the same round with a base that was finalized
+		// in this round or after.
+		// we try to notify if either the round state changed or if we haven't
+		// sent any notification yet (this is to guard against seeing enough
+		// votes to finalize before having precommited)
+		let state_changed = last_state.finalized != new_state.finalized;
+		let sent_finality_notifications = self.best_finalized.is_some();
+
+		if new_state.completable && (state_changed || !sent_finality_notifications) {
```
rphmeier: I see, so this ensures that we always attempt to send notifications until we've precommitted, because when we precommit and the state doesn't change we still need to send the notification.

PR #109, opened 2020-02-06
A warning and a comment

drskalman:
- Added a warning for when the prevote ghost happens not to be the descendent of  the estimate. According to @AlistairStewart this only can happen when there is a massive equivocation so it desrves the extra attention if ever we fell into that case.

- Added a comment to make the code of handling priamry propose more clear following the advice of @andresilva.


drskalman:
[clabot:check]

PR #110, opened 2020-03-02
add more trace loggings

xlc:
Improves https://github.com/paritytech/substrate/issues/4921

This allow me to know which validators are not participating grandpa. It can be hard to know because they are still doing babe fine.

### DIFF
--- a/src/voter/voting_round.rs
+++ b/src/voter/voting_round.rs
@@ -203,6 +203,7 @@ impl<H, N, E: Environment<H, N>> VotingRound<H, N, E> where
 
 		// early exit if the current round is not completable
 		if !self.votes.completable() {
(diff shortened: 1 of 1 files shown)


andresilva:
@xlc I totally get why you are doing this as I have done the same, I actually wrote some code that would log only the authorities that didn't vote. Currently there is no better way to get this information out (hence why we're doing paritytech/substrate#4921), but I think this is too much to be logging by default (e.g. on Kusama we are running all our nodes with trace logging and with 160 validators this would just spam the logs a lot). So I would suggest that we don't merge these changes and instead you just keep them locally for debugging.

xlc:
Make sense. Will do.

ltfschoen:
I think this would have been helpful

PR #114, opened 2020-03-24
Expose voter state

octol:
Implement #55 and support https://github.com/paritytech/substrate/issues/4921 / https://github.com/paritytech/substrate/pull/5375

Expose voter state to the outside world. Useful for querying diagnostics

Left to do: 
- [x] Populate `background_rounds` https://github.com/paritytech/finality-grandpa/pull/114/files#diff-7dbd927baa51274d73fc2a7cb663e166R484

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25,6 +25,7 @@
 //! Equivocation detection and vote-set management is done in the `round` module.
 //! The work for actually casting votes is done in the `voter` module.
 
+#![warn(missing_docs)]
 #![cfg_attr(not(feature = "std"), no_std)]
 
 #[cfg(not(feature = "std"))]
@@ -90,6 +91,7 @@ pub struct Prevote<H, N> {
 }
 
 impl<H, N> Prevote<H, N> {
+	/// Create a new Prevote from the targets block's hash and number.
 	pub fn new(target_hash: H, target_number: N) -> Self {
 		Prevote { target_hash, target_number }
 	}
@@ -106,6 +108,7 @@ pub struct Precommit<H, N> {
 }
 
 impl<H, N> Precommit<H, N> {
+	/// Create a new Precommit from the targets block's hash and number.
 	pub fn new(target_hash: H, target_number: N) -> Self {
 		Precommit { target_hash, target_number }
 	}
@@ -123,14 +126,18 @@ pub struct PrimaryPropose<H, N> {
 }
 
 impl<H, N> PrimaryPropose<H, N> {
+	/// Create a new primary proposed block from the tarets block's hash and
+	/// number.
 	pub fn new(target_hash: H, target_number: N) -> Self {
 		PrimaryPropose { target_hash, target_number }
 	}
 }
 
+/// Top-level Error type
 #[derive(Clone, PartialEq)]
 #[cfg_attr(any(feature = "std", test), derive(Debug))]
 pub enum Error {
+	/// Raised when the block is not a descendent of the given base block.
 	NotDescendent,
 }
 
@@ -226,7 +233,7 @@ pub enum Message<H, N> {
 	/// A precommit message.
 	#[cfg_attr(feature = "derive-codec", codec(index = "1"))]
 	Precommit(Precommit<H, N>),
-	// Primary proposed block.
+	/// Primary proposed block.
 	#[cfg_attr(feature = "derive-codec", codec(index = "2"))]
 	PrimaryPropose(PrimaryPropose<H, N>),
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -91,7 +91,7 @@ pub struct Prevote<H, N> {
 }
 
 impl<H, N> Prevote<H, N> {
-	/// Create a new Prevote from the targets block's hash and number.
+	/// Create a new prevote for the given block (hash and number).
 	pub fn new(target_hash: H, target_number: N) -> Self {
 		Prevote { target_hash, target_number }
 	}
@@ -108,7 +108,7 @@ pub struct Precommit<H, N> {
 }
 
 impl<H, N> Precommit<H, N> {
-	/// Create a new Precommit from the targets block's hash and number.
+	/// Create a new precommit for the given block (hash and number).
 	pub fn new(target_hash: H, target_number: N) -> Self {
 		Precommit { target_hash, target_number }
 	}
@@ -126,14 +126,13 @@ pub struct PrimaryPropose<H, N> {
 }
 
 impl<H, N> PrimaryPropose<H, N> {
-	/// Create a new primary proposed block from the tarets block's hash and
-	/// number.
+	/// Create a new primary proposal for the given block (hash and number).
 	pub fn new(target_hash: H, target_number: N) -> Self {
 		PrimaryPropose { target_hash, target_number }
 	}
 }
 
-/// Top-level Error type
+/// Top-level error type used by this crate.
 #[derive(Clone, PartialEq)]
 #[cfg_attr(any(feature = "std", test), derive(Debug))]
 pub enum Error {
@@ -233,7 +232,7 @@ pub enum Message<H, N> {
 	/// A precommit message.
 	#[cfg_attr(feature = "derive-codec", codec(index = "1"))]
 	Precommit(Precommit<H, N>),
-	/// Primary proposed block.
+	/// A primary proposal message.
 	#[cfg_attr(feature = "derive-codec", codec(index = "2"))]
 	PrimaryPropose(PrimaryPropose<H, N>),
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -136,7 +136,7 @@ impl<H, N> PrimaryPropose<H, N> {
 #[derive(Clone, PartialEq)]
 #[cfg_attr(any(feature = "std", test), derive(Debug))]
 pub enum Error {
-	/// Raised when the block is not a descendent of the given base block.
+	/// The block is not a descendent of the given base block.
 	NotDescendent,
 }
 
--- a/src/testing.rs
+++ b/src/testing.rs
@@ -199,17 +199,16 @@ pub mod environment {
 	}
 
 	impl crate::voter::Environment<&'static str, u32> for Environment {
-		type Timer = Pin<Box<dyn Future<Output = Result<(), Error>> + Unpin + Send + Sync>>;
+		type Timer = Box<dyn Future<Output = Result<(), Error>> + Unpin + Send + Sync>;
 		type Id = Id;
 		type Signature = Signature;
-		type In = Pin<
+		type In =
 			Box<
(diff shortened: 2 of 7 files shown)


octol:
> It looks like @octol hasn't signed our **C**ontributor **L**icense **A**greement, yet.
> 
> > The purpose of a CLA is to ensure that the guardian of a project's outputs has the necessary ownership or grants of rights over all contributions to allow them to distribute under the chosen licence.
> > [Wikipedia](http://en.wikipedia.org/wiki/Contributor_License_Agreement)
> 
> You can read and sign our full Contributor License Agreement at the following URL: https://cla.parity.io
> 
> Once you've signed, please reply to this thread with `[clabot:check]` to prove it.
> 
> Many thanks,
> 
> Parity Technologies CLA Bot

[clabot:check]

src/voter/past_rounds.rs
```diff
@@ -286,6 +290,13 @@ impl<H, N, E: Environment<H, N>> PastRounds<H, N, E> where
 		}
 	}
 
+	pub(super) fn voting_rounds(&self) -> impl Iterator<Item = &VotingRound<H, N, E>> {
```
andresilva, line 293: needs docs

src/voter/mod.rs
```diff
@@ -420,6 +423,76 @@ fn instantiate_last_round<H, N, E: Environment<H, N>>(
 	}
 }
 
+pub trait VoterState<Id> {
```
andresilva, line 426: These and the types below need docs :)

src/testing.rs
```diff
@@ -198,11 +199,19 @@ pub mod environment {
 	}
 
 	impl crate::voter::Environment<&'static str, u32> for Environment {
-		type Timer = BoxFuture<'static, Result<(),Error>>;
+		type Timer = Pin<Box<dyn Future<Output = Result<(), Error>> + Unpin + Send + Sync>>;
```
andresilva, line 202: I need to double check why we need to pin this and the stream/sink below.

octol:
I updated the tests in https://github.com/paritytech/substrate/pull/5375 so that now the check for `background_rounds` should be reliable

src/testing.rs
octol: Removed in latest

src/voter/mod.rs
octol: Good catch!

src/voter/past_rounds.rs
octol: Agree!

src/voter/mod.rs
```diff
@@ -682,26 +796,30 @@ impl<H, N, E: Environment<H, N>, GlobalIn, GlobalOut> Voter<H, N, E, GlobalIn, G
...
 
-		trace!(target: "afg", "Best round at {} has become completable. Starting new best round at {}",
-			self.best_round.round_number(),
-			self.best_round.round_number() + 1,
-		);
+			if !should_start_next { return Poll::Pending }
+
+			trace!(target: "afg", "Best round at {} has become completable. Starting new best round at {}",
+				   inner.best_round.round_number(),
+				   inner.best_round.round_number() + 1,
```
andresilva, line 820:
```suggestion
				inner.best_round.round_number(),
				inner.best_round.round_number() + 1,
```

src/voter/mod.rs
```diff
@@ -1239,17 +1433,49 @@ mod tests {
...
+				report::RoundState::<Id> {
+					total_weight,
+					threshold_weight,
+					prevote_current_weight: VoteWeight(0),
+					prevote_ids: Default::default(),
+					precommit_current_weight: VoteWeight(0),
+					precommit_ids: Default::default(),
+				}
+			)
+		);
```
andresilva, line 1464:
```suggestion
		);
		
```

src/voter/mod.rs
```diff
@@ -1239,17 +1433,49 @@ mod tests {
...
+			Some(&report::RoundState::<Id> {
+				total_weight,
+				threshold_weight,
+				prevote_current_weight: VoteWeight(3),
+				prevote_ids: voter_ids.clone(),
+				precommit_current_weight: VoteWeight(3),
+				precommit_ids: voter_ids,
+			})
+		);
+
```
andresilva, line 1476:
```suggestion
```

src/voter/mod.rs
```diff
@@ -420,6 +423,84 @@ fn instantiate_last_round<H, N, E: Environment<H, N>>(
...
+pub mod report {
+	use std::collections::{HashMap, HashSet};
+	use crate::weights::{VoteWeight, VoterWeight};
+
+	/// Basic data struct for the state of a round.
+	#[derive(PartialEq, Eq, Clone)]
+	#[cfg_attr(test, derive(Debug))]
+	pub struct RoundState<Id: Eq + std::hash::Hash> {
+		pub total_weight: VoterWeight,
+		pub threshold_weight: VoterWeight,
```
rphmeier, line 440: missing docs on these fields, `VoterState` fields, and the `mod report`, and `fn voter_state`. There are even more missing docs - remember the guideline that all public items should be documented.

src/voter/mod.rs
```diff
@@ -420,6 +423,84 @@ fn instantiate_last_round<H, N, E: Environment<H, N>>(
...
+	/// for passing on to other systems.
+	#[derive(PartialEq, Eq)]
+	#[cfg_attr(test, derive(Debug))]
+	pub struct VoterState<Id: Eq + std::hash::Hash> {
+		pub background_rounds: HashMap<u64, RoundState<Id>>,
+		pub best_round: (u64, RoundState<Id>),
+	}
+}
+
+pub struct Inner<H, N, E> where
```
rphmeier, line 459: Exposing a `struct Inner` from `voter.rs` seems a little ambiguous. Likewise, implementation on `Arc<RwLock<...>>` directly seems difficult from a backwards-compatibility perspective.

rphmeier: It's used in `struct Voter` as well, but it's called `inner` although there are other fields. This would not have to be exposed as public with a wrapper struct, meaning it could gain a more descriptive name than `Inner`.

andresilva: The first approach Jon tried used a struct, the problem was that this struct had to be typed on `H, N, E`, and we'd have to propagate these types throughout the substrate codebase, it was especially hairy because types like `Environment` are currently not explicitly typed anywhere but would have to be due to the RPC initialization process. In substrate this trait is used as a trait object to avoid the type proliferation.

rphmeier: OK, I assumed something like that. Still, the API could be improved substantially with a wrapping struct.

octol: Thanks for the feedback @rphmeier :)
I think we can actually remove the `pub` from `Inner` if I'm not mistaken? I'll add the missing docs

andresilva: IMO I think it's better to keep the struct anyway (although not exported) to be able to lock inner as a whole for correctness (rather than best and background distinctly).

octol: Updated the docs and removed the `pub` from `Inner`.

> OK, I assumed something like that. Still, the API could be improved substantially with a wrapping struct.

Sorry, do you mean that the API could have been improved if using a simple struct instead of the trait would have been practical? Or that the API could be improved by creating a struct wrapping `Arc<RwLock<...>>` just so that we don't have to `impl` directly on `Arc<RwLock<...>>`?

I started sketching on the latter but I got stuck on finding a good name ... :/
Probably `VoterState` would be appropriate for such a struct, but then the trait would need to be renamed ... `VoterStateQueryable`? Doesn't roll of the tongue very well ...

andresilva: `struct SharedVoterState<...>(Arc<RwLock<Inner<...>>>);` and then you implement `VoterState` for `SharedVoterState`.

octol: Added the `SharedVoterState` to wrap `Arc<RwLock<Inner>>`

src/voter/mod.rs
```diff
@@ -420,6 +430,102 @@ fn instantiate_last_round<H, N, E: Environment<H, N>>(
 	}
 }
 
+/// Trait for querying the state of the voter. Used by `Voter` to return a queryable object without exposing too may
```
andresilva, line 433:
```suggestion
/// Trait for querying the state of the voter. Used by `Voter` to return a queryable object 
/// without exposing too many data types.
```

src/voter/mod.rs
```diff
@@ -420,6 +430,102 @@ fn instantiate_last_round<H, N, E: Environment<H, N>>(
...
+/// data types.
+pub trait VoterState<Id: Eq + std::hash::Hash> {
+	/// Returns a plain data type, `report::VoterState`, describing the current state of the voter relevant to the
```
andresilva, line 436:
```suggestion
	/// Returns a plain data type, `report::VoterState`, describing the current state 
	/// of the voter relevant to the voting process.
```

src/voter/mod.rs
```diff
@@ -420,6 +430,102 @@ fn instantiate_last_round<H, N, E: Environment<H, N>>(
...
+		/// The identities of nodes that have cast precommits so far.
+		pub precommit_ids: HashSet<Id>,
+	}
+
+	/// Basic data struct for the current state of the voter in a form suitable
+	/// for passing on to other systems.
+	#[derive(PartialEq, Eq)]
+	#[cfg_attr(test, derive(Debug))]
+	pub struct VoterState<Id: Eq + std::hash::Hash> {
+		/// Voting rounds run in the background.
```
andresilva, line 471:
```suggestion
		/// Voting rounds running in the background.
```

src/voter/mod.rs
```diff
@@ -420,6 +430,102 @@ fn instantiate_last_round<H, N, E: Environment<H, N>>(
 	}
 }
 
+/// Trait for querying the state of the voter. Used by `Voter` to return a queryable object without exposing too may
+/// data types.
+pub trait VoterState<Id: Eq + std::hash::Hash> {
+	/// Returns a plain data type, `report::VoterState`, describing the current state of the voter relevant to the
+	/// voting process.
+	fn voter_state(&self) -> report::VoterState<Id>;
```
andresilva, line 438: Maybe we could rename this to `get` otherwise currently it looks like this: `voter.voter_state().voter_state()`.

octol: Updated!

src/lib.rs
```diff
@@ -123,14 +126,17 @@ pub struct PrimaryPropose<H, N> {
...
 	pub fn new(target_hash: H, target_number: N) -> Self {
 		PrimaryPropose { target_hash, target_number }
 	}
 }
 
+/// Top-level error type used by this crate.
 #[derive(Clone, PartialEq)]
 #[cfg_attr(any(feature = "std", test), derive(Debug))]
 pub enum Error {
+	/// Raised when the block is not a descendent of the given base block.
```
JoshOrndorff, line 139:
```suggestion
	/// The block is not a descendent of the given base block.
```

PR #115, opened 2020-04-30
Backport 114: Expose voter state

andresilva:


andresilva:
This is a breaking change unfortunately, requires some `Sync` bounds to be added. Will instead upgrade substrate to v0.12.0.

PR #122, opened 2020-05-08
Fix background round pruning

andresilva:
Whenever we catch-up to a later round we instantiate the caught up round as completed and keep running it in the background. This particular round instance won't have access to the previous round state (since we caught up) and therefore won't be able to vote. Whenever we finalize something in a round we send a notification to the voter but this code assumes that the round will be executed to its terminal state (precommitted) until any finality notification is sent. A consequence of not sending any finality notifications is also that we won't be pruning any background rounds based on this.

In Substrate I tracked a background round leak down to this issue. If the node can't keep up with the latest round (e.g. if it's not an authority gossip data for the latest round takes too long to reach it) and it's keeping up with the latest round through catch up messages then these background rounds would never be pruned.

Additionally I also made it so that when finalizing something through a commit we also prune the background rounds.

### DIFF
--- a/src/voter/voting_round.rs
+++ b/src/voter/voting_round.rs
@@ -657,18 +657,27 @@ impl<H, N, E: Environment<H, N>> VotingRound<H, N, E> where
 		let sent_finality_notifications = self.best_finalized.is_some();
 
 		if new_state.completable && (state_changed || !sent_finality_notifications) {
-			if let (&Some(State::Precommitted), Some((ref f_hash, ref f_number))) = (&self.state, new_state.finalized) {
-				let commit = Commit {
-					target_hash: f_hash.clone(),
-					target_number: *f_number,
-					precommits: self.votes.finalizing_precommits(&*self.env)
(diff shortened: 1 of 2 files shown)


src/voter/voting_round.rs
```diff
@@ -657,18 +657,29 @@ impl<H, N, E: Environment<H, N>> VotingRound<H, N, E> where
...
-				let finalized = (f_hash.clone(), *f_number, self.votes.number(), commit.clone());
-
-				let _ = self.finalized_sender.unbounded_send(finalized);
-				self.best_finalized = Some(commit);
+			let precommitted = matches!(self.state, Some(State::Precommitted));
+			// we only cast votes when we have access to the previous round state,
+			// which won't be the case whenever we catch up to a later round.
+			let cant_vote = self.last_round_state.is_none();
+
+			if precommitted || cant_vote {
```
octol, line 665: I like how introducing these variables makes the logic easier to follow!

PR #128, opened 2021-01-26
update codec to 2.0

thiolliere:
Can we release a new version of finality-grandpa with these change ? This is needed for substrate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,9 +12,9 @@ futures = { version = "0.3.5", default-features = false }
 futures-timer = { version = "3.0.2", optional = true }
 log = { version = "0.4", optional = true }
 parking_lot = { version = "0.11", optional = true }
-parity-scale-codec = { version = "1.3.5", default-features = false, features = ["derive"], optional = true }
+parity-scale-codec = { version = "2.0.0", default-features = false, features = ["derive"], optional = true }
 num = { package = "num-traits", version = "0.2", default-features = false }
-rand = { version = "0.7.0", optional = true }
+rand = { version = "0.8.0", optional = true }
 either = { version = "1.6.1", default-features = false }
 
 [dev-dependencies]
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -12,9 +12,9 @@ futures = { version = "0.3.5", default-features = false }
 futures-timer = { version = "3.0.2", optional = true }
 log = { version = "0.4", optional = true }
 parking_lot = { version = "0.11", optional = true }
-parity-scale-codec = { version = "1.3.5", default-features = false, features = ["derive"], optional = true }
+parity-scale-codec = { version = "2.0.0", default-features = false, features = ["derive"], optional = true }
 num = { package = "num-traits", version = "0.2", default-features = false }
-rand = { version = "0.7.0", optional = true }
+rand = { version = "0.8.0", optional = true }
```
thiolliere, line 17: I just checked that it compiles

src/fuzz_helpers.rs
```diff
@@ -407,7 +407,7 @@ pub fn execute_fuzzed_graph(data: &[u8]) {
 
 	/// A vote-node on the graph.
 	#[derive(Default, Clone, Debug)]
-	struct Vote { prevote: u8, precommit: u8 };
+	struct Vote { prevote: u8, precommit: u8 }
```
thiolliere, line 410: this is just an unnecessary trailing semicolon

andresilva:
@thiolliere Does it change the underlying encoding? Asking because it's easiest to do backport and do a minor release after.

andresilva (approved):

andresilva:
@thiolliere Nevermind, I will just release a new major.

thiolliere:
> @thiolliere Does it change the binary encoding? Asking because it's easiest to backport and do a minor release after.

no it doesn't change any binary encoding, but wouldn't the backport be a breaking change anyway ? as libraries using it would loose the implementation of Encode/Decode for parity-scale-code 1.0 no ?

andresilva:
Yeah, you are right. No worries, I checked and there's nothing new in master besides updated dependencies, everything else was backported already to v0.12, so there's no problem with releasing v0.13.

andresilva:
Merging before CI finishes since Travis is taking forever to run.

PR #129, opened 2021-01-26
Update dependencies

andresilva:
- Updates some dependencies
- Migrates to quickcheck 1.0
- Makes the fuzzing tests work again

Merge after #128.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,8 @@ signals emitted on the underlying blockchain.
 
 ### Fuzzing
 
-To run the fuzzing test harness you need to install either `afl` or `cargo-fuzz`:
+To run the fuzzing test harness you need to install either `afl` or `cargo-fuzz` (you'll need a nightly Rust toolchain
+for this):
 
 ```sh
 cargo install cargo-fuzz
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Add this to your Cargo.toml:
 
 ```toml
 [dependencies]
-finality-grandpa = "0.12"
+finality-grandpa = "0.13"
 ```
 
 **Features:**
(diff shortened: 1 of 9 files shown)


fuzz/afl_targets/graph.rs
```diff
@@ -12,11 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#[macro_use]
+extern crate afl;
```
thiolliere, line 16: I actually tend to prefer not importing macro and just calling them from their path, but I'm fine anyway,

src/bitfield.rs
```diff
@@ -201,21 +207,36 @@ pub struct Bit1 {
...
 	}
 
 	#[test]
 	fn set_bit() {
 		fn prop(mut a: Bitfield, idx: usize) -> bool {
+			// let's bound the max index at 1000 times the number of bits in the
+			// bitfield. this is needed because otherwise when calling `set_bit`
+			// with a large value we might try to allocate more memory than we
+			// have.
+			let idx = idx.min(a.capacity() * 1000);
```
thiolliere, line 238: wouldn't it make more sense to have fixed limit like `2^20` or something?
Anyway it is fine.

thiolliere (approved):
looks good to me, I'm not familiar with the code base but seems legit.

fuzz/afl_targets/graph.rs
andresilva: Me too! But I needed to make this change to make it compile, `fuzz` is calling another macro inside it `__fuzz`.

src/bitfield.rs
andresilva: Yeah good idea.

fuzz/afl_targets/graph.rs
thiolliere: ah I see

andresilva:
Thanks for the review. Updated with your suggestion. It seems that quickcheck's search has gotten smarter in the meantime since some of the edge cases that I had to close now in `Arbitrary` weren't found before :P

README.md
```diff
@@ -73,6 +73,36 @@ Most importantly this crate does not deal with authority set changes. It assumes
...
+```sh
+cargo install cargo-fuzz
+cargo install afl
+```
+
+#### libfuzzer
+
+```sh
+cargo fuzz run graph
+cargo fuzz run round
```
octol, line 89: This might require nightly rustc? Is that worth mentioning, or perhaps a note about it is expected to be obsolete soon anyway?

octol (approved):
Looks good!

README.md
andresilva: Good point, I only use nightly so didn't notice it. Updated the readme.

rphmeier (approved):

PR #134, opened 2021-06-11
vote-graph: remove duplicate base in ancestry iter

zhiqiangxu:
`ancestry` already contains `self.base`, so no need to chain it.


andresilva (approved):
Thanks!

PR #136, opened 2021-06-15
remove RwLock to drop Sync bounds

andresilva:
This PR replaces a `RwLock` with a `Mutex` in order to remove all the existing `Sync` bounds. This should ease integration with the rest of the async/await ecosystem (e.g. async_trait). The `RwLock` is not exposed is almost never contended so in practice this shouldn't make any difference.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "finality-grandpa"
-version = "0.14.0"
+version = "0.14.1"
 description = "PBFT-based finality gadget for blockchains"
 authors = ["Parity Technologies <admin@parity.io>"]
 license = "Apache-2.0"
--- a/src/voter/mod.rs
+++ b/src/voter/mod.rs
@@ -30,7 +30,7 @@ use futures::channel::mpsc::{self, UnboundedReceiver};
 #[cfg(feature = "std")]
 use log::trace;
 
-use parking_lot::RwLock;
+use parking_lot::Mutex;
 
 use std::collections::VecDeque;
 use std::pin::Pin;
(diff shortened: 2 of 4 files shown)


octol (approved):

andresilva:
Still fighting with some bounds on substrate so don't merge this yet.

src/voter/mod.rs
```diff
@@ -61,9 +61,9 @@ pub trait Environment<H: Eq, N: BlockNumberOps>: Chain<H, N> {
 	type Timer: Future<Output = Result<(), Self::Error>> + Unpin;
 	/// Associated future type for the environment used when asynchronously computing the
 	/// best chain to vote on. See also [`Self::best_chain_containing`].
-	type BestChain: Future<Output = Result<Option<(H, N)>, Self::Error>> + Send + Sync + Unpin;
+	type BestChain: Future<Output = Result<Option<(H, N)>, Self::Error>> + Send + Unpin;
 	/// The associated Id for the Environment.
-	type Id: Ord + Clone + Eq + ::std::fmt::Debug;
+	type Id: Clone + Eq + Ord + Hash + std::fmt::Debug;
```
rphmeier, line 66: `Hash` bound addition here requires a major version bump

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "finality-grandpa"
-version = "0.14.0"
+version = "0.14.1"
```
rphmeier, line 3:
```suggestion
version = "0.15.0"
```

rphmeier (approved):
LGTM other than nit

src/voter/mod.rs
andresilva: I reverted this change as there's no need to add it here (kept it in the `voter_state` method).

PR #140, opened 2021-08-29
format grandpa

gilescope:
Been reading up on grandpa code and have a few PRs, but first time to format grandpa.

**No code changes**, only cargo fmt.

Grandpa this won't hurt at all.


gilescope:
The exception is that `Cargo.toml` was autoformatted by the vscode "even better toml" plugin.

adoerr (approved):

PR #141, opened 2021-08-29
order field not needed

gilescope:
Unless I am mistaken, the VoterSet.voters is immutable and already ordered by Id as it's a BTreeMap. 
So there appears to be no need to maintain the order separately to this?

### DIFF
--- a/src/voter_set.rs
+++ b/src/voter_set.rs
@@ -20,7 +20,6 @@ use crate::{
 	std::{
 		collections::{btree_map::Entry, BTreeMap},
 		num::{NonZeroU64, NonZeroUsize},
-		vec::Vec,
 	},
 	weights::VoterWeight,
 };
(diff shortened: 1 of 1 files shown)


gilescope:
(I checked that the `nth()` of `order.iter()` and `nth()` of `voters.iter()` returned the same order (At the time I wasn't sure if the order might have been reversed, but it wasn't).

andresilva (approved):
I think the reason we had the `order` field there is because initially voters where being stored in an `HashMap`, not sure why we didn't notice this when changing to `BTreeMap` since we were just taking the keys from the map as-is (without explicitly sorting). Good catch.

andresilva:
The only thing I can come up with for this is that with the previous code `nth` would be `O(log(n))` since it's a `O(1)` call to the `Vec` to find the voter key and then `O(log(n))` search on the BTreeMap. With the current code it will be `O(n)` as we will iterate through the `BTreeMap` to find the `nth` member. This will be called multiple times every time we process a new vote (since we need to figure out the voter weight).

gilescope:
Good point - I want to make things strictly better. So at the moment it's O(log(n)) whether we look up by Id or by nth()?

If we keep order but change our storage around a little and store the index in the BTreeMap instead of the VoterInfo like this:
```
	voters: BTreeMap<Id, usize>,
	order: Vec<VoterInfo>,
```
then we get O(1) for the nth() lookups and the by Id becomes O(1) + O(log(n))  - that seems a good trade?

I can try that in a different PR and we can see what it looks like?

andresilva:
Sounds good to me, I think both `get` and `nth` will eventually be called whenever a new vote is imported. Btw, I also think we're probably splitting hairs here and this doesn't really matter in practice as the number of validators (and hence the size of those data structures) will likely always be bounded enough that `O(n)` vs `O(log(n))` doesn't make a difference :p but when reviewing the PR I was thinking if there was any reason why the `order` field was added and came up with that.

gilescope:
Closing in favor of #143

PR #143, opened 2021-09-08
nth() now O(1), get still O(log(n))

gilescope:
I saw that we could still get down to one data structure and improve the nth() complexity to O(1) by using binary search on a vec sorted by Ids. We don't really need a BTreeMap as the structure is immutable - ordered inserts aren't needed once constructed.

The sort is like tim-sort i.e. very efficient if inputs happen to be already sorted by Id. Hopefully this is strictly better than what we had before in all dimensions.

### DIFF
--- a/src/voter_set.rs
+++ b/src/voter_set.rs
@@ -18,7 +18,6 @@
 
 use crate::{
 	std::{
-		collections::{btree_map::Entry, BTreeMap},
 		num::{NonZeroU64, NonZeroUsize},
 		vec::Vec,
 	},
@@ -32,10 +31,8 @@ use crate::{
 /// equipped with a total order, given by the ordering of the voter's IDs.
 #[derive(Clone, PartialEq, Eq, Debug)]
 pub struct VoterSet<Id: Eq + Ord> {
-	/// The voters in the set.
-	voters: BTreeMap<Id, VoterInfo>,
-	/// The total order associated with the keys in `voters`.
-	order: Vec<Id>,
+	/// The total ordered voters set
+	voters: Vec<(Id, VoterInfo)>,
 	/// The required weight threshold for supermajority w.r.t. this set.
 	threshold: VoterWeight,
 	/// The total weight of all voters.
@@ -58,9 +55,9 @@ impl<Id: Eq + Ord> VoterSet<Id> {
 		I: IntoIterator<Item = (Id, u64)>,
 	{
 		let weights = weights.into_iter();
+		let mut voters = Vec::with_capacity(weights.size_hint().0);
 
 		// Populate the voter set, thereby calculating the total weight.
-		let mut voters = BTreeMap::new();
 		let mut total_weight = 0u64;
 		for (id, weight) in weights {
 			if let Some(w) = NonZeroU64::new(weight) {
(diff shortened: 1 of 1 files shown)


src/voter_set.rs
```diff
@@ -92,46 +77,70 @@ impl<Id: Eq + Ord> VoterSet<Id> {
...
+				current_weight = 0;
+			}
 		}
 
-		let threshold = threshold(total_weight);
+		// Duplicate ids and weights after this position have already been processed.
+		voters.truncate(position);
 
-		Some(VoterSet { voters, order, total_weight, threshold })
+		Some(VoterSet { voters, total_weight, threshold: threshold(total_weight) })
```
andresilva, line 109: Can we keep the original code for aggregating with a `BTreeMap` and then do a simple conversion to `Vec<(Id, VoterInfo)>`? The code will be simpler and the voter set constructor isn't run that often so we don't really need to optimize it.

src/voter_set.rs
```diff
@@ -92,46 +77,70 @@ impl<Id: Eq + Ord> VoterSet<Id> {
...
 			// SAFETY: By VoterSet::new()
-			NonZeroUsize::new_unchecked(self.order.len())
+			NonZeroUsize::new_unchecked(self.voters.len())
 		}
 	}
 
 	/// Whether the set contains a voter with the given ID.
 	pub fn contains(&self, id: &Id) -> bool {
-		self.voters.contains_key(id)
+		self.voters.binary_search_by(|(an_id, _)| an_id.cmp(id)).is_ok()
```
andresilva, line 130:
```suggestion
		self.voters.binary_search_by_key(&id, |(id, _)| id).is_ok()
```

src/voter_set.rs
```diff
@@ -92,46 +77,70 @@ impl<Id: Eq + Ord> VoterSet<Id> {
...
 
-		Some(VoterSet { voters, order, total_weight, threshold })
+		Some(VoterSet { voters, total_weight, threshold: threshold(total_weight) })
 	}
 
 	/// Get the voter info for the voter with the given ID, if any.
 	pub fn get(&self, id: &Id) -> Option<&VoterInfo> {
-		self.voters.get(id)
+		self.voters
+			.binary_search_by(|(an_id, _)| an_id.cmp(id))
```
andresilva, line 115:
```suggestion
			.binary_search_by_key(&id, |(id, _)| id)
```

andresilva (changes requested):

src/voter_set.rs
```diff
@@ -57,10 +55,9 @@ impl<Id: Eq + Ord> VoterSet<Id> {
 		Id: Ord + Clone,
 		I: IntoIterator<Item = (Id, u64)>,
 	{
-		let weights = weights.into_iter();
```
gilescope, line 60: it's an into_iter already isn't it?

src/voter_set.rs
gilescope: Better?

src/voter_set.rs
andresilva: `weights` is something that can be converted into an iterator but we need to call `into_iter()` on it to get an iterator, I think the code won't compile if you remove this line.

andresilva: Nevermind, it should work now since you're not calling `size_hint` anymore and just using it in the for loop below.

andresilva (approved):

PR #145, opened 2021-09-16
Enable dependabot and cargo-deny

andresilva:


deny.toml
```diff
@@ -0,0 +1,203 @@
+# This template contains all of the possible sections and their default values
```
andresilva, line 1: This was generated with `cargo deny init` and the only thing I changed was the list of allowed licenses below.

adoerr (approved):

PR #146, opened 2021-09-16
Fix clippy warnings

andresilva:
Just minor things to make the CI check green :)

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,0 +1,1 @@
+type-complexity-threshold = 9000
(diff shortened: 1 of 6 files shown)


.clippy.toml
```diff
@@ -0,0 +1 @@
+type-complexity-threshold = 9000
```
andresilva, line 1: I don't want this lint to run and this is the easiest way to disable it without adding clippy tags to the source code (which I don't want to do).

adoerr (approved):

andresilva:
Thanks!

gilescope (commented):
LGTM

PR #147, opened 2021-11-14
Move to rust 2021

NukeManDan:


andresilva:
That was probably caused by some compiler bug. I tested now and didn't have any issues, only needed to make a small change in a test.

andresilva (approved):

PR #150, opened 2022-02-06
Prepare release 0.14.4 with SCALE 3.0.0

wigy-opensource-developer:


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "finality-grandpa"
-version = "0.14.4"
+version = "0.14.5"
```
andresilva, line 3: I think this should be a major bump to "0.15.0" as people depending on `Encode/Decode` of these types will also need to bump their scale codec version.

andresilva (changes requested):
Do we also need to (or should?) bump scale-info to 2.0?

Cargo.toml
andresilva: Cool, I think we can do all of that in this PR :)

andresilva (approved):

PR #152, opened 2022-04-26
Fix commit validation to allow precommits lower than commit target

andresilva:
This might be needed in a situation where there is an equivocation as part of the commit, which is treated as "voting for all blocks", and can therefore allow for finalizing a block higher than those precommits target. I added tests to cover these scenarios.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -597,6 +597,8 @@ impl<H, N, S, Id> HistoricalVotes<H, N, S, Id> {
 
 #[cfg(test)]
 mod tests {
+	use super::*;
+	use crate::testing::chain::{DummyChain, GENESIS_HASH};
 
 	#[cfg(feature = "derive-codec")]
 	#[test]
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -449,28 +441,37 @@ where
...
+		.min_by_key(|precommit| precommit.target_number)
+		.map(|precommit| (precommit.target_hash.clone(), precommit.target_number))
+	{
+		None => return Ok(validation_result),
+		Some(base) => base,
+	};
+
+	// check that all precommits are for blocks that are equal to or descendants
+	// of the round base
+	let all_precommits_higher_than_base = commit.precommits.iter().all(|signed| {
```
davxy, line 460: Is the number of invalid voters ( in order to initialize `num_invalid_voters`) deliberately ignored?
Currently is left to 0 (default).
I don't think is paramount, and indeed looks like the value is just ignored by Substrate

davxy:
Not strictly related to this PR modifications.

Isn't this instruction preventing to count all the equivocations?
https://github.com/paritytech/finality-grandpa/blob/47555597fd5056f1aea5c1c7777246308670b57f/src/lib.rs#L481-L484

If a second equivocation from the same `id` is found, then here we're eagerly exit the function thus we skip counting other potential equivocations from other validators.

Again, looks like these counters are currently not used by Substrate... so maybe that is not so important.

src/lib.rs
andresilva: Currently the semantics for `num_invalid_voters` is votes from voters that are not part of the validator set, i.e. a message signed by some random key, so it should be OK to stay at 0 here (we didn't check so we don't know).

andresilva:
> Not strictly related to this PR modifications.
> 
> Isn't this instruction preventing to count all the equivocations?
> 
> https://github.com/paritytech/finality-grandpa/blob/47555597fd5056f1aea5c1c7777246308670b57f/src/lib.rs#L481-L484
> 
> If a second equivocation from the same `id` is found, then here we're eagerly exit the function thus we skip counting other potential equivocations from other validators.

The idea was to bail as soon as possible since at this point we know the `Commit` is invalid, there might indeed be some more equivocations in the commit but I think we don't need to care anymore at this point.

davxy (approved):

Issue #155, opened 2022-10-20
Make `Equivocation` accept two rounds


vmarkushin:
It's possible to have different rounds in two different justifications signed by the same authority, but [`Equivocation`](https://github.com/paritytech/finality-grandpa/blob/eb26b0c86dbb6949435b627f85eceb026f96ebe9/src/lib.rs#L210) structure only accepts one round, which makes it impossible to report misbehaviour in such case. This problem is especially acute in light clients that verify GRANDPA since there is no way to track each round in the client.

seunlanlege:
How should light client respond to when there's a supermajority equivocation for a differennt round, but the same target?

vmarkushin:
@andresilva thank you for the explanation. Makes sense.

PR #156, opened 2022-10-20
Make `Equivocation` have separate rounds for each vote

vmarkushin:
solves: #155


andresilva (changes requested):
I don't think this is correct and I think instead we should do the validation I explained in #156 regarding always ensuring the invariant that whatever gets finalized is consistent with what we had previously observed as finalized, and shutting down in case that invariant is broken. The fact that you have seen two conflicting votes by a given validator is not by itself an equivocation, you can't assert that this voter is bad as it might have been duped into voting that way by the dishonest majority. This could even happen in a scenario where there aren't even byzantine voters at all e.g. on round 1 everyone votes for block #A while I vote for block #A', at the end of the round block #A gets finalized and I re-org to this fork and on the next round I vote for block #B (which is a descendant of #A and "conflicts" with my previous vote on #A').

PR #157, opened 2022-11-10
Fix typo in panic message

timwu20:
Was looking at the `VoteGraph` code and noticed a typo in the `introduce_branch` function.


andresilva (approved):
Thanks! Sorry for taking so long.

PR #168, opened 2025-06-03
voter: Stop the voter logic on terminated global input stream

lexnv:
This PR ensures that the voter logic stops when the `global_in` stream is terminated.

While at it, have replaced the `log` dependency with `tracing`

### DIFF
--- a/src/voter/voting_round.rs
+++ b/src/voter/voting_round.rs
@@ -18,7 +18,7 @@
 use futures::ready;
 use futures::{channel::mpsc::UnboundedSender, prelude::*};
 #[cfg(feature = "std")]
-use log::{debug, trace, warn};
+use tracing::{debug, trace, warn};
 
 use std::{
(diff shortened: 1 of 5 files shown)


src/voter/voting_round.rs
```diff
@@ -716,10 +727,11 @@ where
 								last_round_estimate.0
 							}
 						},
-						Err(crate::Error::NotDescendent) => {
+						Err(err) => {
 							// This is only possible in case of massive equivocation
 							warn!(
 								target: LOG_TARGET,
+								?err,
 								"Possible case of massive equivocation: \
```
rphmeier: probably don't want this log line to trigger if it's a stream-closed error.

andresilva (approved):
overall lgtm, just need to handle rob's comment (only log the equivocation statement if the error is `NotDescendent`)

andresilva (approved):

PR #169, opened 2025-06-06
voter: Ensure prevoting future is polled directly after advancing the state

lexnv:
This PR avoids a call to the `cx.waker().wake_by_ref()` by polling the appropriate `finish_prevoting` future.

Previously, we relied on the context waker to wake up the whole async task if the state advanced to `Prevoting` via the `start_prevoting` fn. I believe the intended behavior was to save the waker on the subsequent call by polling the `best_chain` future.

Instead, we poll the `best_chain` future directly without waking up the task. 

cc @paritytech/networking @alexggh https://github.com/paritytech/polkadot-sdk/issues/8543

## Next Steps
- [ ] This needs a bit of testing to ensure that `wake_by_ref` did not have unexpected side-effects which unblocked other futures from continuing to operate, which may explain the finality stall in the linked issue

### DIFF
--- a/src/voter/voting_round.rs
+++ b/src/voter/voting_round.rs
@@ -538,13 +538,6 @@ where
 					);
 
 					let (base, best_chain) = this.construct_prevote(last_round_state);
-
-					// since we haven't polled the future above yet we need to
-					// manually schedule the current task to be awoken so the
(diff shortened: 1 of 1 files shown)


andresilva (commented):
I honestly no longer remember any of this :-(

